PR #7, opened 2017-05-02
Expose methods on container models

kevensen:
Depends On:
ManageIQ/manageiq-providers-openshift#2
ManageIQ/manageiq#14798

https://bugzilla.redhat.com/show_bug.cgi?id=1440141

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_container_route.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_container_route.rb
@@ -1,5 +1,7 @@
 module MiqAeMethodService
   class MiqAeServiceContainerRoute < MiqAeServiceModelBase
+    require_relative 'mixins/miq_ae_service_container_resource_mixin'
+    include MiqAeServiceContainerResourceMixin
     expose :ext_management_system,    :association => true
     expose :container_project,        :association => true
     expose :container_service,        :association => true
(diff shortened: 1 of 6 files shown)


gmcculloug:
@simon3z Can you please review this from the container side.  At a quick glance I'm a little concerned that we are exposing too much here.

kevensen:
@gmcculloug I've updated the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1440141) to better describe what I'm hoping to accomplish.  I am absolutely open to any recommendations for improvement.

lib/miq_automation_engine/service_models/miq_ae_service_container_route.rb
```diff
@@ -8,5 +8,9 @@ class MiqAeServiceContainerRoute < MiqAeServiceModelBase
     expose :labels,                   :association => true
     expose :is_tagged_with?
     expose :tags
+    expose :spec
```
mkanoor: @kevensen 
Can the spec, spec=,annotations, tidy_provider_definition be moved into a mixin and then we can include it in 3 files

kevensen: Yes they can.  I'll do that.

simon3z:
@kevensen can you provide the other PR dependencies for this one? I guess we'll need the backend for these before and there may be changes (e.g. IIRC we're planning to rename `add_role_to_user`, etc.)

kevensen:
**Depends On:**
https://github.com/ManageIQ/manageiq-providers-openshift/pull/2
https://github.com/ManageIQ/manageiq/pull/14798

gmcculloug:
Moved dependences from comment above (https://github.com/ManageIQ/manageiq-automation_engine/pull/7#issuecomment-302399718) to the PR description so they are easy to identify.

gmcculloug:
@kevensen Looks like the dependent PRs are stale.  Is this still being worked on or should we close for now and re-open if there is need?

kevensen:
@gmcculloug Let’s go ahead and close this and the related PR’s for now.

PR #8, opened 2017-05-03
Expose evacuate VM method to automate

andyvesel:
Expose the `evacuate` method for instances in OpenStack Cloud Provider

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb
@@ -41,6 +41,7 @@ module MiqAeMethodService
     expose :compliances,           :association => true
     expose :last_compliance,       :association => true
     expose :ems_events,            :association => true
+    expose :evacuate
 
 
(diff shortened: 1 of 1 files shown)


lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb
```diff
@@ -41,6 +41,7 @@ class MiqAeServiceVmOrTemplate < MiqAeServiceModelBase
     expose :compliances,           :association => true
     expose :last_compliance,       :association => true
     expose :ems_events,            :association => true
+    expose :evacuate,              :override_return => nil
```
mkanoor: @andyvesel 
Why are we overriding the return value? Can't the caller just ignore the returned value.
This gives us problems when we starting implementing these via the REST API.

gmcculloug: @andyvesel I think the idea is to not use `:override_return` and allow the data to pass back to the caller.  Any idea what the return value would be here?

andyvesel: @gmcculloug @mkanoor  `:override_return` has been currently removed

PR #9, opened 2017-05-03
Expose delete method for Cloud Subnet

andyvesel:
This exposes the delete method for Cloud Subnet in OpenStack Cloud Provider.

Original PR https://github.com/ManageIQ/manageiq/pull/14516
Depends on ManageIQ/manageiq#15087

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_cloud_subnet.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_cloud_subnet.rb
@@ -3,5 +3,6 @@ module MiqAeMethodService
     expose :cloud_network,     :association => true
     expose :availability_zone, :association => true
     expose :vms,               :association => true
+    expose :delete_cloud_subnet
   end
 end


lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb
```diff
@@ -41,6 +41,7 @@ class MiqAeServiceVmOrTemplate < MiqAeServiceModelBase
     expose :compliances,           :association => true
     expose :last_compliance,       :association => true
     expose :ems_events,            :association => true
+    expose :evacuate,              :override_return => nil
```
mkanoor: @andyvesel Why is the return being overridden here, can't the caller from the Automate Method just ignore it? This will cause us issues when we start implementing this in REST API.

gmcculloug:
@andyvesel @Ladas I cannot find a `delete_cloud_subnet` method defined on this object, so this will not work.  Currently there is only `delete_cloud_subnet_queue` and `raw_delete_cloud_subnet`.

andyvesel:
@gmcculloug yes, right. I've just created a PR with those methods for a Cloud Subnet model in a main repo.
https://github.com/ManageIQ/manageiq/pull/15087

PR #10, opened 2017-05-04
Remove zip-zip

bdunne:
Split from https://github.com/ManageIQ/manageiq/pull/14881

From what I can tell, the only thing holding us back from dropping our dependency on zip-zip was the usage of some aliased constants, so I changed to use the ones provided by Rubyzip.


bdunne:
@mkanoor Please review
@miq-bot add_label fine/no

PR #16, opened 2017-05-05
Added support for embedded methods

mkanoor:
https://bugzilla.redhat.com/show_bug.cgi?id=1418374

This optional backend feature allows you to embed Automate Methods into other Automate Methods.

There is no UI support for it yet.

We collect all the embedded method bodies into a single script before execution. If the methods throws any errors we correctly log the line number from the offending methods.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -8,11 +8,11 @@ module MiqAeEngine
       @num_methods = 0
     end
 
-    def with_server(inputs, bodies, method_name, line_hash)
+    def with_server(inputs, bodies, method_name, script_info)
       setup if num_methods == 0
       self.num_methods += 1
       svc = MiqAeMethodService::MiqAeService.new(@workspace, inputs)
-      yield build_method_content(bodies, method_name, svc.object_id, line_hash)
+      yield build_method_content(bodies, method_name, svc.object_id, script_info)
     ensure
       svc.destroy # Reset inputs to empty to avoid storing object references
       self.num_methods -= 1
@@ -67,23 +67,23 @@ module MiqAeEngine
 
     # code building
 
-    def build_method_content(bodies, method_name, miq_ae_service_token, line_hash)
+    def build_method_content(bodies, method_name, miq_ae_service_token, script_info)
       [
-        dynamic_preamble(method_name, miq_ae_service_token, line_hash),
+        dynamic_preamble(method_name, miq_ae_service_token, script_info),
         RUBY_METHOD_PREAMBLE,
-        bodies.flatten,
+        bodies,
         RUBY_METHOD_POSTSCRIPT
-      ].join("\n")
+      ].flatten.join("\n")
     end
 
-    def dynamic_preamble(method_name, miq_ae_service_token, line_hash)
-      line_hash_yaml = line_hash.to_yaml
+    def dynamic_preamble(method_name, miq_ae_service_token, script_info)
+      script_info_yaml = script_info.to_yaml
       <<-RUBY.chomp
 MIQ_URI = '#{drb_server.uri}'
 MIQ_ID = #{miq_ae_service_token}
 RUBY_METHOD_NAME = '#{method_name}'
-LINE_HASH_YAML = '#{line_hash_yaml}'
-RUBY_METHOD_PREAMBLE_LINES = #{RUBY_METHOD_PREAMBLE_LINES + 5 + line_hash_yaml.lines.count}
+SCRIPT_INFO_YAML = '#{script_info_yaml}'
+RUBY_METHOD_PREAMBLE_LINES = #{RUBY_METHOD_PREAMBLE_LINES + 5 + script_info_yaml.lines.count}
 RUBY
     end
 
@@ -112,7 +112,6 @@ begin
 
   DRbObject.send(:undef_method, :inspect)
   DRbObject.send(:undef_method, :id) if DRbObject.respond_to?(:id)
-  LINE_HASH = YAML.load(LINE_HASH_YAML)
 
   DRb.start_service
   $evmdrb = DRbObject.new_with_uri(MIQ_URI)
@@ -153,10 +152,11 @@ class Exception
   end
 
   def get_file_info(line)
-    LINE_HASH.each do |fqname, item|
-      return fqname, line - item[:start] if line.between?(item[:start], item[:end])
+    script_info = YAML.load(SCRIPT_INFO_YAML)
+    script_info.each do |fqname, range|
+      return fqname, line - range.begin if range.cover?(line)
     end
-    return "?", line
+    return RUBY_METHOD_NAME, line
   end
 
   alias backtrace_without_evm backtrace
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -8,11 +8,11 @@ module MiqAeEngine
       @num_methods = 0
     end
 
-    def with_server(inputs, body, method_name)
+    def with_server(inputs, bodies, method_name, line_hash)
       setup if num_methods == 0
       self.num_methods += 1
       svc = MiqAeMethodService::MiqAeService.new(@workspace, inputs)
-      yield build_method_content(body, method_name, svc.object_id)
+      yield build_method_content(bodies, method_name, svc.object_id, line_hash)
     ensure
       svc.destroy # Reset inputs to empty to avoid storing object references
       self.num_methods -= 1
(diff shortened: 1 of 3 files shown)


mkanoor:
@Fryguy please review

mkanoor:
Some of the initial prototype work was done in PR https://github.com/ManageIQ/manageiq/pull/14286 which used method parameters to embed methods. We have abandoned that approach now that we have a separate column to store the name of embedded methods.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -110,6 +112,7 @@ class AutomateMethodException < StandardError
 
   DRbObject.send(:undef_method, :inspect)
   DRbObject.send(:undef_method, :id) if DRbObject.respond_to?(:id)
+  LINE_HASH = YAML.load(LINE_HASH_YAML)
```
Fryguy: I don't think we need to pay the penalty for this on every invocation...it can probably just be called directly in `get_file_info` when needed.  I would expect this to only happen during Exceptions, which themselves would be rare.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -148,6 +152,13 @@ def backtrace_with_evm
     value ? filter_backtrace(value) : value
   end
 
+  def get_file_info(line)
+    LINE_HASH.each do |fqname, item|
+      return fqname, line - item[:start] if line.between?(item[:start], item[:end])
+    end
+    return "?", line
```
Fryguy: Previously, this would be RUBY_METHOD_NAME, so I think that would be preferable than `?`

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
...
+    def self.get_bodies_and_line_numbers(obj, aem)
+      embeds = get_embeds(obj.workspace, aem) << {:data => aem.data, :fqname => aem.fqname}
+      code_start = 0
+      line_hash = {}
+      bodies = []
+      embeds.each do |item|
+        bodies << item[:data]
+        line_count = item[:data].lines.count
+        code_end   = code_start + line_count
+        line_hash[item[:fqname]] = {:start => code_start, :end => code_end}
```
Fryguy: Instead of this hash, you can use a Range object.  Then later your between can actually use the Range#include? or Range#cover? method.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -67,21 +67,23 @@ def drb_cache_timeout
...
-    def build_method_content(body, method_name, miq_ae_service_token)
+    def build_method_content(bodies, method_name, miq_ae_service_token, line_hash)
       [
-        dynamic_preamble(method_name, miq_ae_service_token),
+        dynamic_preamble(method_name, miq_ae_service_token, line_hash),
         RUBY_METHOD_PREAMBLE,
-        body,
+        bodies.flatten,
         RUBY_METHOD_POSTSCRIPT
       ].join("\n")
```
Fryguy: I feel like this should be changed to [].flatten.join("\n") instead of the flatten *inside* the array.  I know it works both ways, but I feel it's more intuitive this way.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -67,21 +67,23 @@ def drb_cache_timeout
...
     end
 
-    def dynamic_preamble(method_name, miq_ae_service_token)
+    def dynamic_preamble(method_name, miq_ae_service_token, line_hash)
```
Fryguy: I'm not a fan of `line_hash` as a variable name...trying to think of something better.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
...
+    private_class_method :get_bodies_and_line_numbers
+
+    def self.get_embeds(workspace, method_obj)
+      method_obj.embedded_methods.collect do |name|
+        method_name, klass, ns = get_embed_method_name(name)
+        match_ns = workspace.overlay_method(ns, klass, method_name)
+        cls = ::MiqAeClass.find_by_fqname("#{match_ns}/#{klass}")
+        aem = ::MiqAeMethod.find_by(:class_id => cls.id, :name => method_name) if cls
+        raise MiqAeException::MethodNotFound, "Embedded method #{name} not found" unless aem
+        fqname = "#{match_ns}/#{klass}/#{method_name}"
```
Fryguy: Does MiqAeMethod not have a find_by_fqname?  If so, would that make sense to have?

Fryguy: If it did, I think it would consolidate the last four lines into:

```ruby
aem = ::MiqAeMethod.find_by_fqname("#{match_ns}/#{klass}/#{method_name}")
raise MiqAeException::MethodNotFound, "Embedded method #{name} not found" unless aem
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
       threads.each(&:exit)
     end
     private_class_method :cleanup
+
+    def self.get_bodies_and_line_numbers(obj, aem)
```
Fryguy: Generally, `get_` as a method prefix is a smell.  it's typically not necessary, or shows that the method name could be more expressive.  In this case `bodies_and_line_numbers` is sufficient or perhaps `bodies_and_line_numbers_from_object`.

Fryguy: Same goes for the get_embeds and get_embed_method_name below.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
...
+        line_count = item[:data].lines.count
+        code_end   = code_start + line_count
+        line_hash[item[:fqname]] = {:start => code_start, :end => code_end}
+        code_start = code_end + 1
+      end
+      return bodies, line_hash
+    end
+    private_class_method :get_bodies_and_line_numbers
+
+    def self.get_embeds(workspace, method_obj)
```
Fryguy: since the method in method_obj is `embedded_methods` I feel this method should also say `embedded_methods` somewhere as well.  Abbreviating to `embeds` only confuses the reader because it's a new term where one already exists (and where the original term is much clearer IMO).

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
mkanoor: how about script_info?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
mkanoor: @Fryguy All the find_by_fqname calls are expensive, since it is a computed value and we have to walk up the tree from instance -> class -> namespace(s) -> domain

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
...
     private_class_method :cleanup
+
+    def self.get_bodies_and_line_numbers(obj, aem)
+      embeds = get_embeds(obj.workspace, aem) << {:data => aem.data, :fqname => aem.fqname}
+      code_start = 0
+      line_hash = {}
+      bodies = []
+      embeds.each do |item|
+        bodies << item[:data]
+        line_count = item[:data].lines.count
```
mkanoor: @Fryguy won't that be treated as a single line.

spec/engine/miq_ae_method_spec.rb
```diff
@@ -2,18 +2,22 @@
...
+        end
         # rubocop:enable Style/SingleLineMethods, Style/EmptyLineBetweenDefs
       end.new
     end
 
-    let(:aem)    { double("AEM", :data => script, :fqname => "/my/automate/method") }
+    let(:aem)    { double("AEM", :data => script, :fqname => "/my/automate/method", :embedded_methods => embeds) }
     let(:obj)    { double("OBJ", :workspace => workspace) }
-    let(:inputs) { [] }
+    let(:inputs) { {} }
```
gmcculloug: Looks like `inputs` being an array has been wrong for a long time.  👍

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -212,5 +213,42 @@ def self.cleanup(method_pid, threads)
...
+    end
+    private_class_method :bodies_and_line_numbers
+
+    def self.embedded_methods(workspace, method_obj)
+      method_obj.embedded_methods.collect do |name|
+        method_name, klass, ns = embedded_method_name(name)
+        match_ns = workspace.overlay_method(ns, klass, method_name)
+        cls = ::MiqAeClass.find_by_fqname("#{match_ns}/#{klass}")
+        aem = ::MiqAeMethod.find_by(:class_id => cls.id, :name => method_name) if cls
+        raise MiqAeException::MethodNotFound, "Embedded method #{name} not found" unless aem
```
gmcculloug: Was looking specifically for this missing embedded method check and corresponding tests.  Looks good.

Issue #20, opened 2017-05-10
Resolve "warning: already initialized constant" message in tests


gmcculloug:
```
/home/travis/build/ManageIQ/manageiq-automation_engine/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb:115: warning: already initialized constant MiqAeMethodService::MiqAeServiceManageIQ_Providers_InfraManager
/home/travis/build/ManageIQ/manageiq-automation_engine/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb:115: warning: previous definition of MiqAeServiceManageIQ_Providers_InfraManager was here
```

Narrowed down the errors to test "loads all mapped models" here: https://github.com/ManageIQ/manageiq-automation_engine/blob/4fcb6fe274b71d072039e9f45448dd60bbe3237f/spec/miq_ae_service_spec.rb#L88

Related to the `LEGACY_MODEL_NAMES` mapping here:
https://github.com/ManageIQ/manageiq-automation_engine/blob/4fcb6fe274b71d072039e9f45448dd60bbe3237f/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_legacy.rb#L9

gmcculloug:
Fixed in PR #22

PR #21, opened 2017-05-11
Change Openstack_InfraManager service model use real model for subclass

gmcculloug:
Openstack_InfraManager service model was using the [alias EmsInfra](https://github.com/ManageIQ/manageiq/blob/master/app/models/aliases/ems_infra.rb) instead of the real subclass `ManageIQ::Providers::InfraManager`

Fixes issues #20


gmcculloug:
This PR is replaced by PR #22 since this one did not originate from my fork of this repo.

PR #27, opened 2017-05-23
[Depends on ManageIQ/manageiq#15192] Expose update method for Cloud Subnet

andyvesel:
This exposes the update method for Cloud Subnet in OpenStack Cloud Provider.
Depends on https://github.com/ManageIQ/manageiq/pull/15192


gmcculloug:
@andyvesel Please review this and dependent PR to determine if they are still needed.

PR #28, opened 2017-05-24
Changed MiqQueue.put calls to use method in mixin

mkanoor:
Instead of using MiqQueue.put directly use sync_or_async_ems_operation

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
@@ -36,7 +36,7 @@ module MiqAeMethodService
     METHODS_WITH_NO_ARGS = %w(scan)
     METHODS_WITH_NO_ARGS.each do |m|
       define_method(m) do
-        sync_or_async_ems_operation(false, m, [])
+        sync_or_async_ems_operation(false, m)
         true
       end
     end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
@@ -2,6 +2,8 @@ module MiqAeMethodService
   class MiqAeServiceHost < MiqAeServiceModelBase
     require_relative "mixins/miq_ae_service_custom_attribute_mixin"
     include MiqAeServiceCustomAttributeMixin
+    require_relative "mixins/miq_ae_service_ems_operations_mixin"
+    include MiqAeServiceEmsOperationsMixin
 
     expose :storages,              :association => true
     expose :read_only_storages
@@ -34,16 +36,8 @@ module MiqAeMethodService
     METHODS_WITH_NO_ARGS = %w(scan)
     METHODS_WITH_NO_ARGS.each do |m|
       define_method(m) do
-        ar_method do
-          MiqQueue.put(
-            :class_name  => @object.class.name,
-            :instance_id => @object.id,
-            :method_name => m,
-            :zone        => @object.my_zone,
-            :role        => "ems_operations"
-          )
-          true
-        end
+        sync_or_async_ems_operation(false, m, [])
+        true
       end
     end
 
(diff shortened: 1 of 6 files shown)


mkanoor:
@gmcculloug @tinaafitz Please review

lib/miq_automation_engine/service_models/miq_ae_service_host.rb
```diff
@@ -34,16 +36,8 @@ class MiqAeServiceHost < MiqAeServiceModelBase
...
-          MiqQueue.put(
-            :class_name  => @object.class.name,
-            :instance_id => @object.id,
-            :method_name => m,
-            :zone        => @object.my_zone,
-            :role        => "ems_operations"
-          )
-          true
-        end
+        sync_or_async_ems_operation(false, m, [])
```
gmcculloug: I think you mean `nil` instead of `[]` for the `args` parameter.

lib/miq_automation_engine/service_models/miq_ae_service_host.rb
```diff
@@ -65,14 +59,7 @@ def ems_custom_get(key)
...
     def ems_custom_set(attribute, value)
-      MiqQueue.put(
-        :class_name  => @object.class.name,
-        :instance_id => @object.id,
-        :method_name => 'set_custom_field',
-        :zone        => @object.my_zone,
-        :role        => 'ems_operations',
-        :args        => [attribute, value]
-      ) if @object.is_vmware?
+      sync_or_async_ems_operation(false, 'set_custom_field', [attribute, value]) if @object.is_vmware?
```
gmcculloug: I would suggest moving this into a VMware subclassed service model and drop the `if @object.is_vmware?` check.

lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-vmware-infra_manager-host_esx.rb
```diff
@@ -1,119 +1,49 @@
...
-          :class_name  => @object.class.name,
-          :instance_id => @object.id,
-          :method_name => "vim_shutdown",
-          :zone        => @object.my_zone,
-          :role        => "ems_operations",
-          :args        => [force]
-        ) if @object.is_vmware?
-        true
-      end
+      sync_or_async_ems_operation(false, "vim_shutdown", [force]) if @object.is_vmware?
```
gmcculloug: This is already a VMware model so `if @object.is_vmware?` can be dropped from all these methods.

lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb
```diff
@@ -47,16 +47,8 @@ class MiqAeServiceVmOrTemplate < MiqAeServiceModelBase
...
-          MiqQueue.put(
-            :class_name  => @object.class.name,
-            :instance_id => @object.id,
-            :method_name => m,
-            :zone        => @object.my_zone,
-            :role        => "ems_operations"
-          )
-          true
-        end
+        sync_or_async_ems_operation(false, m, [])
```
gmcculloug: Again thinking you mean `nil` instead of `[]`.

mkanoor:
@gmcculloug Args is serialized as an Array by MiqQueue.
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_queue.rb#L72

kbrock:
This is looking very nice @mkanoor 
Would be nice if we didn't support synchronous (since it will make it harder to separate the direct db access from models), but this is a good step forward

lib/miq_automation_engine/service_models/miq_ae_service_host.rb
gmcculloug: This method is currently implemented in the base `Host` model and is checking for `is_vmware?` so there is no need to do the check in the service model. (See https://github.com/ManageIQ/manageiq/blob/87061966261086b377c85cf09302938ab7be3029/app/models/host.rb#L1264)

As a followup PR this should be changed to follow the implementation in `vm_or_template.rb` to use the `raw_*` pattern.

For now I would just suggest removing the `is_vmware?` check.

mkanoor:
@kbrock 
These aren't direct DB calls, they are actions on the resource which can only be performed on the EMS that this resource belongs to. e.g stop/start/scan

PR #29, opened 2017-05-25
remove global variable _miq_worker_current_msg

kbrock:
Trying to get away from using `$_miq_worker_current_msg` global variable

This will get us closer to using any queuing system.

/cc @mkanoor

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -24,19 +24,7 @@ module MiqAeEngine
       :msg_timeout => 60.minutes
     }.merge(options)
 
-    static_keys = [:class_name, :method_name, :zone, :msg_timeout]
-
-    ##############################################
-    # IF there is no current message
-    # OR any of the static keys in the current message do NOT match the value in the options for the same key,
-    # THEN create a new message
-    # ELSE requeue the current message with the new options
-    ##############################################
-    if $_miq_worker_current_msg.nil? || options.any? { |key, value| static_keys.include?(key) ? ($_miq_worker_current_msg.send(key) != value) : false }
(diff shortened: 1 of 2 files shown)


kbrock:
/cc @Fryguy @gmcculloug @jrafanie do you see an issue with this approach?
Is there a reason we need to access this global to `requeue`?

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -24,19 +24,7 @@ def self.deliver_queue(args, options = {})
...
-    # IF there is no current message
-    # OR any of the static keys in the current message do NOT match the value in the options for the same key,
-    # THEN create a new message
-    # ELSE requeue the current message with the new options
-    ##############################################
-    if $_miq_worker_current_msg.nil? || options.any? { |key, value| static_keys.include?(key) ? ($_miq_worker_current_msg.send(key) != value) : false }
-      MiqQueue.put(options)
-    else
-      $_miq_worker_current_msg.requeue(options)
-    end
```
jrafanie: @mkanoor is it ok to drop the notion of requeueing of automate messages?

mkanoor: @jrafanie I think we can delete it, its some kind of an optimization that might have been added to prevent adding too many queue entries

Fryguy:
My only concern is what we are losing with removing requeue.  What does the requeue method differently than MiqQueue.put?

kbrock:
The current code says to `requeue` if all the parameters are identical.

`MiqQueue#requeue` will filter out any fields that don't want to be inserted.
The only field that may have special behavior is the reuse of `$_miq_worker_current_msg.task_id`. But I think those should be the same.

PR #32, opened 2017-06-01
[WIP] Add MiqAeService model for service template catalogs

fdupont-redhat:
Currently, it is not possible to explore service catalogs from Automate, while it is to explore service templates. This pull request closes the gap.

This is my first pull request ever, so please be gentle and guide me :)

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_catalog.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_catalog.rb
@@ -1,0 +1,7 @@
+module MiqAeMethodService
+  class MiqAeServiceServiceTemplateCatalog < MiqAeServiceModelBase
+    expose :service_template_catalogs, :association => true
+    expose :service_templates,         :association => true
+    expose :tenant,                    :association => true
+  end
+end
--- a/spec/service_models/miq_ae_service_service_template_catalog_spec.rb
+++ b/spec/service_models/miq_ae_service_service_template_catalog_spec.rb
@@ -5,18 +5,18 @@ module MiqAeServiceServiceTemplateCatalogSpec
         Spec::Support::MiqAutomateHelper.create_service_model_method('SPEC_DOMAIN', 'EVM', 'AUTOMATE', 'test1', 'test')
         @ae_method     = ::MiqAeMethod.first
         @ae_result_key = 'foo'
-        @service_template   = FactoryGirl.create(:service_template_catalog)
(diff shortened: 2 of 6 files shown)


lib/miq_automation_engine/service_models/miq_ae_service_service_template_catalog.rb
```diff
@@ -0,0 +1,7 @@
+module MiqAeMethodService
+  class MiqAeServiceServiceTemplateCatalog < MiqAeServiceModelBase
+    expose :service_template_catalogs, :association => true
```
gmcculloug: This is not an exposed relationship on the rails side: https://github.com/ManageIQ/manageiq/blob/master/app/models/service_template_catalog.rb#L6-L7

This line just needs to be removed.

spec/service_models/miq_ae_service_service_template_catalog_spec.rb
```diff
@@ -0,0 +1,32 @@
+module MiqAeServiceServiceTemplateCatalogSpec
+  describe MiqAeMethodService::MiqAeServiceServiceTemplateCatalog do
+    context "through an automation method" do
```
gmcculloug: This context is not needed, should be removed.

fdupont-redhat: Would you remove the whole block or just move the code out of it ?

lib/miq_automation_engine/service_models/miq_ae_service_service_template_catalog.rb
fdupont-redhat: Just to be sure, you talk about the line above ? The circular relationship ?

gmcculloug: Yes, the `service_template_catalogs` relationship does not exist so the `expose` needs to be removed.

spec/service_models/miq_ae_service_service_template_catalog_spec.rb
gmcculloug: I meant the whole `context "through an automation method" do` block should be removed.

fdupont-redhat:
I created a new PR, based on a branch, without the code from other PRs and cleaned from the non existent relationship and the useless test context. Here it is: https://github.com/ManageIQ/manageiq-automation_engine/pull/40.

gmcculloug:
Closed in favor of PR https://github.com/ManageIQ/manageiq-automation_engine/pull/40

PR #34, opened 2017-06-02
Move create_service_provision_request into MiqAeServiceMethods for consistency

### LINKED ISSUE #33: Use of the new create_service_provision_request method is inconsistent with other create_*_request methods
The new ```create_service_provision_request``` method is called directly from ```$evm```, however the 2 'sibling' methods ```create_automation_request``` and ```create_provision_request``` are called via ```$evm.execute```. This is inconsistent and potentially confusing.

Suggest we move ```create_service_provision_request``` into ```MiqAeServiceMethods``` for consistency.

fdupont-redhat:
Fixes #33 
https://bugzilla.redhat.com/show_bug.cgi?id=1457754

Related to:
https://github.com/ManageIQ/manageiq-content/pull/126

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
@@ -110,6 +110,11 @@ module MiqAeMethodService
       MiqAeServiceModelBase.wrap_results(AutomationRequest.create_request(options, user, auto_approve))
     end
 
+    def self.create_service_provision_request(svc_template, options = nil)
+      result = ar_object(svc_template).provision_request(@workspace.ae_user, options)
+      MiqAeServiceModelBase.wrap_results(result)
+    end
+
     def self.drb_undumped(klass)
       _log.info "Entered: klass=#{klass.name}"
       klass.include(DRbUndumped) unless klass.ancestors.include?(DRbUndumped)
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
@@ -111,7 +111,7 @@ module MiqAeMethodService
     end
 
     def self.create_service_provision_request(svc_template, options = nil)
-      result = ar_object(svc_template).provision_request(@workspace.ae_user, options)
+      result = svc_template.object_send(:provision_request, User.current_user, options)
       MiqAeServiceModelBase.wrap_results(result)
     end
 
--- a/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
@@ -182,7 +182,8 @@ module MiqAeServiceMethodsSpec
 
       it "create service request" do
         allow(workspace).to receive(:disable_rbac)
-        expect_any_instance_of(ServiceTemplate).to receive(:provision_request).with(user, svc_options).and_return(miq_request)
+        allow(ServiceTemplate).to receive(:find).with(template.id).and_return(template)
+        expect(template).to receive(:provision_request).with(user, svc_options).and_return(miq_request)
 
         result = miq_ae_service.execute(:create_service_provision_request, svc_template, svc_options)
         expect(result).to be_kind_of(MiqAeMethodService::MiqAeServiceMiqRequest)
--- a/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
@@ -162,5 +162,32 @@ module MiqAeServiceMethodsSpec
       ct.reload
       expect(ct.entries.collect(&:name).include?('fred')).to be_truthy
     end
+
+    context "#create_service_provision_request" do
+      let(:options) { {:fred => :flintstone} }
+      let(:svc_options) { {:dialog_style => "medium"} }
+      let(:user) { FactoryGirl.create(:user_with_group) }
+      let(:template) { FactoryGirl.create(:service_template_ansible_playbook) }
+      let(:miq_request) { FactoryGirl.create(:service_template_provision_request) }
+      let(:svc_template) do
+        MiqAeMethodService::MiqAeServiceServiceTemplate.find(template.id)
+      end
+      let(:workspace) do
+        double("MiqAeEngine::MiqAeWorkspaceRuntime",
+               :root               => options,
+               :persist_state_hash => {},
+               :ae_user            => user)
+      end
+      let(:miq_ae_service) { MiqAeService.new(workspace) }
+
+      it "create service request" do
+        allow(workspace).to receive(:disable_rbac)
+        expect(svc_template).to receive(:instance_variable_get).with('@object').and_return(template)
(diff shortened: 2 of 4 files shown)


gmcculloug:
@fdupont-redhat Thanks for the PR and for referencing the BZ.  

A few quick comments.

- There is no need to create a git issue to link the PR to, a PR is essentially a git issue so you are just duplicating your efforts.
- If you do have a git issue using `Fixes #nn` will automatically close the issue when the PR is merged.  (I updated the description.)

In this case I would say the description you provided in the git issue should really be the description of this PR.  Again, I updated the description to reflect that.

- Will need to modify the existing caller of this method: https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb#L19

I still need to look into failing tests.

fdupont-redhat:
@gmcculloug Thanks for these advices.
For the file to modify in manageiq-content, how do you handle this cross linkage?

gmcculloug:
Create a PR in the content repo and in the title and description you should include the dependency.

Here is an example of the title: https://github.com/ManageIQ/manageiq/pull/13049
And description: https://github.com/ManageIQ/manageiq/pull/15251

Once the new PR is created I also like to update the description in this PR for easy reference.

spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
```diff
@@ -162,5 +162,32 @@ def category_create_script
...
+               :ae_user            => user)
+      end
+      let(:miq_ae_service) { MiqAeService.new(workspace) }
+
+      it "create service request" do
+        allow(workspace).to receive(:disable_rbac)
+        expect(svc_template).to receive(:instance_variable_get).with('@object').and_return(template)
+        expect(template).to receive(:provision_request).with(user, svc_options).and_return(miq_request)
+
+        result = miq_ae_service.execute(:create_service_provision_reques, svc_template, svc_options)
```
mkanoor: @fdupont-redhat 
Is this missing a 't' create_service_provision_request

gmcculloug: Good catch @mkanoor.  Any thoughts on why the test is failing?

fdupont-redhat: @mkanoor, aka. Hawk Eye ;)
I corrected it. It seems the problem comes from the fact that MiqAeMethodService class is not know in this context. I also tried to change the test to:

```ruby
    it "#create_service_provision_request" do
      options = {:fred => :flinstone}
      svc_options = {:dialog_style => "medium"}
      template = FactoryGirl.create(:service_template_ansible_playbook)
      miq_request = FactoryGirl.create(:service_template_provision_request)
      svc_template = MiqAeMethodService::MiqAeServiceServiceTemplate.find(template.id)
      workspace = double("MiqAeEngine::MiqAeWorkspaceRuntime", :root => options, :persist_state_hash => {}, :ae_user => @user)
      #miq_ae_service = MiqAeMethodService.new(workspace)

      allow(workspace).to receive(:disable_rbac)
      expect(svc_template).to receive(:instance_variable_get).with('@object').and_return(template)
      expect(template).to receive(:service_provision_request).with(@user, svc_options).and_return(miq_request)

      result = invoke_ae.execute(:create_service_provision_request, svc_template.inspect, svc_options.inspect)
      expect(result).to be_kind_of(MiqAeMethodService::MiqAeServiceMiqRequest)
    end
```

But invoke_ae doesn't really give ```$evm``` object and tells me

```ruby
undefined method `execute' for #<MiqAeEngine::MiqAeWorkspaceRuntime:0x0000000e3db518>
```

fdupont-redhat: That is probably why there is no test for ```create_automation_request``` and ```create_provision_request```. Should we move these methods to ```lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb``` ? That would also be consistent with ```$evm.create_notification```. And it would be easier to write tests. The drawback is that it would probably require more changes to ```manageiq-content``` repository.

What do you think @gmcculloug, @mkanoor ?

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -110,6 +110,11 @@ def self.create_automation_request(options, userid = "admin", auto_approve = fal
       MiqAeServiceModelBase.wrap_results(AutomationRequest.create_request(options, user, auto_approve))
     end
 
+    def self.create_service_provision_request(svc_template, options = nil)
+      result = ar_object(svc_template).provision_request(@workspace.ae_user, options)
+      MiqAeServiceModelBase.wrap_results(result)
+    end
```
gmcculloug: This method has to change a bit more now that it is running under a different class.  Since we do not have access to `@workspace` we can get the user from the `User.current_user` set but the calling method `execute` [here](https://github.com/ManageIQ/manageiq-automation_engine/blob/2c4b6e9a4813ba142d528b403acdbd1f7916d193/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb#L168).

```ruby
def self.create_service_provision_request(svc_template, options = nil)
  result = svc_template.object_send(:provision_request, User.current_user, options)
  MiqAeServiceModelBase.wrap_results(result)
end
```

spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
gmcculloug: This should get things moving for the test:

```ruby
it "create service request" do
  allow(workspace).to receive(:disable_rbac)
  expect_any_instance_of(ServiceTemplate).to receive(:provision_request).with(user, svc_options).and_return(miq_request)

  result = miq_ae_service.execute(:create_service_provision_request, svc_template, svc_options)
  expect(result).to be_kind_of(MiqAeMethodService::MiqAeServiceMiqRequest)
end
```

fdupont-redhat:
@gmcculloug you rock!

gmcculloug:
@fdupont-redhat Let's make one last change to avoid the rubocop warning about using `expect_any_instance_of`.

I used that call initially because the service model loads it's own copy of the `service_template` and the `expect(template).to receive(:provision_request)` would be called on the wrong copy.

To avoid using `expect_any_instance_of` we just need to return the same template instance when the service model for `ServiceTemplate` performs a `find` to look up the instance. 

Here is my diff that changes the test to use the same instance of service_template:
```diff
-        expect_any_instance_of(ServiceTemplate).to receive(:provision_request).with(user, svc_options).and_return(miq_request)
+        allow(ServiceTemplate).to receive(:find).with(template.id).and_return(template)
+        expect(template).to receive(:provision_request).with(user, svc_options).and_return(miq_request)
```

mkanoor:
@gmcculloug @fdupont-redhat 👍

simaishi:
Fine backport (to manageiq repo) details:

```
$ git log -1
commit ebd69905757ba17d786876ebc18c93fd8df6cda4
Date:   Mon Jun 12 09:44:46 2017 -0400

    Merge pull request #34 from fdupont-redhat/bz1447754
    
    Move create_service_provision_request into MiqAeServiceMethods for consistency
    (cherry picked from commit f650a66f1241de078ffdca508b0e77afb8eb5460)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1461144
```

PR #36, opened 2017-06-05
Add dependency on rubyzip gem

chessbyte:
To enable https://github.com/ManageIQ/manageiq-gems-pending/pull/191


gmcculloug:
@chessbyte @jrafanie Hakiri is recommending to use version 1.2.1.  With the `~>` I know that version will get pulled in anyway, but does it make sense to change it here to clear the security warning?

jrafanie:
@gmcculloug I think so.  Let's use `~>1.2.1`

chessbyte:
@gmcculloug @jrafanie Updated to `~>1.2.1`

gmcculloug:
👍  Will merge when tests pass.

PR #41, opened 2017-06-13
Expose ServiceTemplateContainerTemplate to automate.

lfu:
Expose ServiceTemplateContainerTemplate to automate.

Depends on [manageiq #15356](https://github.com/ManageIQ/manageiq/pull/15356).

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_container_template.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_container_template.rb
@@ -1,0 +1,6 @@
+module MiqAeMethodService
+  class MiqAeServiceServiceTemplateContainerTemplate < MiqAeServiceServiceTemplateGeneric
+    expose :container_template
+    expose :container_manager
+  end
+end


lfu:
cc @bzwei

lib/miq_automation_engine/service_models/miq_ae_service_service_template_container_template.rb
```diff
@@ -0,0 +1,5 @@
+module MiqAeMethodService
+  class MiqAeServiceServiceTemplateContainerTemplate < MiqAeServiceServiceTemplateGeneric
+    expose :container_template
```
bzwei: do we see the need to use `container_template` in the automate? If not immediately, let's hold off this PR because the ae model can be automately generated now.

lfu: @bzwei The dynamic method AvailableProjects uses it:
```
projects = service.try(:container_template).try(:ext_management_system).try(:container_projects)
```

PR #42, opened 2017-06-27
Fix git import's attribute verify_ssl

paulslaby:
When automate model is refreshed from git, attribute verify_ssl should be used.

### DIFF
--- a/app/models/miq_ae_git_import.rb
+++ b/app/models/miq_ae_git_import.rb
@@ -55,6 +55,7 @@ class MiqAeGitImport
     @options['ref'] ||= DEFAULT_BRANCH
     @options['ref_type'] ||= BRANCH
     @options['ref_type'] = @options['ref_type'].downcase
+    @options['verify_ssl'] = @git_repo.verify_ssl
   end
 
   def branch_or_tag
--- a/spec/models/miq_ae_git_import_spec.rb
+++ b/spec/models/miq_ae_git_import_spec.rb
@@ -56,9 +56,10 @@ describe MiqAeGitImport do
         end
 
         let(:import_options) do
-          options.merge("ref_type" => MiqAeGitImport::BRANCH,
-                        "branch"   => branch_name,
-                        "git_dir"  => repo.directory_name)
+          options.merge("ref_type"   => MiqAeGitImport::BRANCH,
+                        "branch"     => branch_name,
+                        "git_dir"    => repo.directory_name,
+                        "verify_ssl" => 1)
         end
 
         before do
@@ -84,10 +85,11 @@ describe MiqAeGitImport do
             basic_options.merge('ref' => branch_name.downcase, 'git_repository_id' => repo.id)
           end
           let(:import_options) do
-            options.merge("ref_type" => MiqAeGitImport::BRANCH,
-                          "ref"      => branch_name,
-                          "branch"   => branch_name,
-                          "git_dir"  => repo.directory_name)
+            options.merge("ref_type"   => MiqAeGitImport::BRANCH,
+                          "ref"        => branch_name,
+                          "branch"     => branch_name,
+                          "verify_ssl" => 1,
+                          "git_dir"    => repo.directory_name)
           end
           it_behaves_like "#import that has a valid branch"
         end
@@ -104,6 +106,7 @@ describe MiqAeGitImport do
         let(:options) do
           basic_options.reverse_merge('ref'               => tag_name,
                                       'git_repository_id' => repo.id,
+                                      'verify_ssl'        => 1,
                                       'ref_type'          => MiqAeGitImport::TAG)
         end
 
@@ -133,7 +136,8 @@ describe MiqAeGitImport do
           let(:options) do
             basic_options.merge('ref'               => tag_name.downcase,
                                 'git_repository_id' => repo.id,
-                                'ref_type'          => MiqAeGitImport::TAG)
+                                'ref_type'          => MiqAeGitImport::TAG,
+                                'verify_ssl'        => 1)
           end
 
           let(:import_options) do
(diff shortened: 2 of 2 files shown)


gmcculloug:
bump @mkanoor

paulslaby:
@mkanoor [This](https://github.com/paulslaby/manageiq-automation_engine/blob/9c31eafda809ad706fd5a2005cde4a438bd4d73a/app/models/miq_ae_git_import.rb#L47) was always nil before the change.

mkanoor:
@paulslaby 
When we create a repository for the first time we should be setting up all the properties (verify_ssl, url, userid, password etc).
https://github.com/paulslaby/manageiq-automation_engine/blob/9c31eafda809ad706fd5a2005cde4a438bd4d73a/app/models/miq_ae_git_import.rb#L26
We call create_repo only if the repository has not been created.
In your case do you already have the repo created or are you creating a new one.

paulslaby:
But the information from the model is used nowhere. That's what I did.


Dne 07.07.17 v 17:40 Madhu Kanoor napsal(a):

mkanoor:
@paulslaby 
https://github.com/ManageIQ/manageiq/blob/master/app/models/git_repository.rb#L105

We add a certificate check callback if verify_ssl is disabled.

paulslaby:
@mkanoor It is nice, but it obviously doesn't work. Have you even tried? 

When you call API refresh domain, not only it is always 1, but it even sets GitRepository.verify_ssl to 1. I don't remember the details and I don't want to debug it again, it stole me and my company many hours. I fixed it and I feel it is good to give it to others. If you think, there is a better fix, do it by yourself. I'm okay with that. I only need it working.

vaclavmiller:
@mkanoor hello, please may you let us know about the progress? Will you accept PR or choose other approach? If so, what is your time estimation when will be resolved?

martinpovolny:
I dare to sum up:

You guus create a GIT repo from the UI. 

(Seems the UI code does not use the  https://github.com/paulslaby/manageiq-automation_engine/blob/master/app/models/miq_ae_git_import.rb#L45
but instead
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/services/git_repository_service.rb which seems wrong but irrelevant)

Then am import goes well for the first time. Right?

Then later, you decide that you need an import and ask for an import through the API (API refresh domain).

At this point things go south. Not only the refresh fails, but in the database on the GitRepository instance/record the `verify_ssl` attribute is set to '1'.

Is that correct?

paulslaby:
@martinpovolny Exactly!

martinpovolny:
So the problem is here:

From the API the UI code is called:
https://github.com/ManageIQ/manageiq/blob/master/app/controllers/api/automate_domains_controller.rb#L19

```
          task_id = GitBasedDomainImportService.new.queue_refresh_and_import(domain.git_repository.url,  
                                                                             ref,                        
                                                                             ref_type,                   
                                                                             current_tenant.id) 
```

That calls `MiqAeDomain.import_git_url`  that calls `MiqAeGitImport.new(options).import` with the options (not including `verify_ssl`).

From there we get to https://github.com/ManageIQ/manageiq-automation_engine/blob/master/app/models/miq_ae_git_import.rb#L26 with `@git_repo` nil and call the https://github.com/ManageIQ/manageiq-automation_engine/blob/master/app/models/miq_ae_git_import.rb#L45 which rewrites `verify_ssl`

`@git_repo.update_attributes(:verify_ssl => @options['verify_ssl'] || OpenSSL::SSL::VERIFY_PEER)`

because `@options['verify_ssl']` is nil.



My summary is that as a quick fix the `GitBasedDomainImportService.new.queue_refresh_and_import(` call should include the `verify_ssl`.

But there's quite some missconception in how the API and UI code use the `app/models/miq_ae_git_import.rb`.

This clearly is a bug. Thank you, @paulslaby for the effort spent on fixing it! 

However the fix in this PR fixes in in a way way that should not be merged :-( 

Ping @abellotti, @mkanoor, @eclarizio : I found your names on the problematic lines. I thinks there needs to be some short discussion on how the code should be rearranged and fixed in all the 3 repositories involved (manageiq, manageiq-ui-classic and manageiq-automation_engine).

martinpovolny:
@paulslaby: please, is there a BZ for this problem? Can you link it? And if not, can you, please, create a BZ? 

It will help a lot with backporting the fixes to the appropriate releases. Thx!

mkanoor:
The problem is in the find_or_create_by which doesn't tell us if the repository exists or we have to create a new one.
https://github.com/ManageIQ/manageiq-automation_engine/blob/master/app/models/miq_ae_git_import.rb#L46

The line after this blindly sets verify_ssl as if its a new record and since the caller didn't pass one we default to 1.

Maybe the solution is to break the find_or_create_by into a find and a separate create call which will preserve the attributes in the object instead.

mkanoor:
I think the find_or_create_by takes a block the second line below needs to be in a block.

mkanoor:
``` ruby
@git_repo = GitRepository.find_or_create_by(:url => @options['git_url']) do |repo|
    repo.update_attributes(:verify_ssl => @options['verify_ssl'] || OpenSSL::SSL::VERIFY_PEER)
end
```
The userid and password could change in the future so we have to allow for them to be updated separately.

martinpovolny:
@mkanoor : I think that your last comment includes a valid way to quick fix the problem.

But what I see is that the API code uses UI code to schedule the refresh/import. That is clearly wrong.

gmcculloug:
Closing in favor of PR https://github.com/ManageIQ/manageiq-automation_engine/pull/52

@martinpovolny Please open an issue for the based on your comments above. 
 (https://github.com/ManageIQ/manageiq-automation_engine/pull/42#issuecomment-315097104)

paulslaby:
Thanks. Bug is here: https://bugzilla.redhat.com/show_bug.cgi?id=1470738

PR #49, opened 2017-07-10
Support for expression methods

mkanoor:
Expression Methods
https://www.pivotaltracker.com/n/projects/1613499/stories/148625717

Expression Methods allow you to use Advance Search Filters as Automate Methods, substituting the user_input from Automate Objects. The advantages of this approach is that the methods run directly in the worker and don't have the overhead of forking a DRb process to run the Automate Methods.

Related PR https://github.com/ManageIQ/manageiq-automation_engine/pull/50

# Steps
1. Create a new Automate Method and set the Location to expression
2. Select the Expression object to use for the expression builder from the drop down list
3. Create an expression using the expression editor controls and any values that need to be set at runtime from automate, mark then as user input. These are positional arguments.
4. In the input parameters section  create arg1,arg2,arg3... for each of the user input fields from Step 3. When specifying the default value for arg1,arg2,... you can use substitution. e.g. _${/#vm_suffix}_

5. Create a field called attributes with the list of attributes that you would like to collect from the search objects. e.g (name, id)

## Input Parameters
* **arg1** The first argument in the expression
* **arg2** The second argument in the expression
* **argn** The nth argument in the expression

### Optional Input Parameters
* **attributes** A comma delimited list of attributes to select from the resultant objects
or
* **distinct** A comma delimited list of attributes which are distinct in the resultant objects

If attributes and distinct are not specified we try to store the result in a variable called
**values** with a hash consisting of **id** and **name**. This makes it compatible with our existing dynamic dialog result set.

* **result_obj** The object where the result data should be stored (default: **current object**)
* **result_attr** The name of the attribute which stores the result (default: **values**)
* **result_type** The result type hash or array (default: **dialog_hash** which matches to our dynamic dialog hash. Valid values are 'hash', 'dialog_hash', 'array', 'simple'
* **on_empty** The method behavior when the search returns an empty list
warn | error | abort (default: **error**)
* **default** The default value in case the result is empty and you select warn

* UI PR https://github.com/ManageIQ/manageiq-ui-classic/pull/1668
* ManageIQ PR: https://github.com/ManageIQ/manageiq/pull/15537

### The Expression Method in Edit Mode

screen shot 2017-07-11 at 2 55 32 pm

### The Expression Method in Display Mode

screen shot 2017-07-11 at 2 54 58 pm

### Dynamic Dialog Instance

The instance below shows how to setup for Dynamic Dialogs, the default properties can be set in the instance and the method will fill in the **values** field. In the examples below the AvailableVMsInternal is an expression method.

screen shot 2017-07-11 at 2 51 21 pm

### DIFF
--- a/spec/miq_ae_expression_method_spec.rb
+++ b/spec/miq_ae_expression_method_spec.rb
@@ -1,0 +1,197 @@
+module MiqAeExpressionMethodSpec
+  include MiqAeEngine
+  describe MiqAeExpressionMethod do
+    include Spec::Support::AutomationHelper
+    let(:user) { FactoryGirl.create(:user_with_group) }
+    let(:vm1) { FactoryGirl.create(:vm, :name => 'test_2.1', :cpu_shares => 400) }
+    let(:vm2) { FactoryGirl.create(:vm, :name => 'test_3.1', :cpu_shares => 400) }
+    let(:vm3) { FactoryGirl.create(:vm, :name => 'test_4.2', :cpu_shares => 11) }
+    let(:complex_qs_exp) do
+      {"and" => [{"STARTS WITH" => {"field" => "Vm-name", "value" => :user_input}},
+                 {"ENDS WITH"   => {"field" => "Vm-name", "value" => :user_input}}]
+      }
+    end
+
+    let(:m_params) do
+      {'arg1'       => {'datatype' => 'string', 'default_value' => 'test'},
+       'arg2'       => {'datatype' => 'string', 'default_value' => '1'},
+       'attributes' => {'datatype' => 'array',  'default_value' => 'name'}
+      }
+    end
+
+    let(:vm_search) do
+      {:db         => 'Vm',
+       :expression => complex_qs_exp}.to_yaml
+    end
+
+    before do
+      allow(User).to receive(:server_timezone).and_return("UTC")
+    end
+
+    it "expression_method" do
+      vm1
+      vm2
+      vm3
+      m_params['result_type'] = {'datatype' => 'string', 'default_value' => 'array'}
+      create_ae_model_with_method(:ae_namespace => 'GAULS',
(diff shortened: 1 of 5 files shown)


mkanoor:
@gmcculloug Travis failure is pending for this PR https://github.com/ManageIQ/manageiq/pull/15537 to get merged.

spec/miq_ae_expression_method_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+    let(:vm_search) do
+      {:db         => 'Vm',
+       :expression => complex_qs_exp}.to_yaml
+    end
+
+    before do
+      allow(User).to receive(:server_timezone).and_return("UTC")
+    end
+
+    it "expression_method" do
```
gmcculloug: Suggest moving these tests into something like a `describe "expression_method" do` block to better group and reduce the duplication in each test title.

spec/miq_ae_expression_method_spec.rb
```diff
@@ -0,0 +1,169 @@
+module MiqAeExpressionMethodSpec
+  include MiqAeEngine
+  describe MiqAeExpressionMethod do
```
gmcculloug: Can we come up with a way to test this class without having to build an automate model and instantiate it each time?  Seems very expensive to run each test and it would be better to focus the tests on the class.  Thoughts?

gmcculloug:
@mkanoor The PR description has great documentation for the feature but I think it could use a few updates/enhancements that would be really useful.

1. There are no longer any `Required Input Parameters`.  Should rename section to `Input Parameters`.
2. For `Optional Input Parameters` the key `distinct` is a tough one to wrap my head around.  An example in the description would be helpful.
3. `result_type` lists the default as `hash` but the code is showing `dialog_hash`
4. `result_type` should list the available options.

PR #50, opened 2017-07-11
Export Import of Expression Methods

mkanoor:
This PR allows expressions to be exported and imported. The expression files are stored with the extension .expr in the export/import deck

https://www.pivotaltracker.com/n/projects/1613499/stories/148625717

### DIFF
--- a/app/models/miq_ae_yaml_export.rb
+++ b/app/models/miq_ae_yaml_export.rb
@@ -168,7 +168,9 @@ class MiqAeYamlExport
     class_obj.ae_methods.sort_by(&:fqname).each do |meth_obj|
       export_file_hash['created_on'] = meth_obj.created_on
       export_file_hash['updated_on'] = meth_obj.updated_on
-      write_method_file(meth_obj, export_file_hash) unless meth_obj.location == 'builtin'
+      if meth_obj.location == 'inline'
+        write_method_file(meth_obj, export_file_hash)
+      end
       write_method_attributes(meth_obj, export_file_hash)
     end
   end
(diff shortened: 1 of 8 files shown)


app/models/miq_ae_yaml_export.rb
```diff
@@ -201,6 +201,8 @@ def get_method_filename(method_obj)
       "#{method_obj.name}#{FILE_EXTENSIONS[method_obj.language]}"
     elsif method_obj['location'] == 'uri'
       "#{method_obj.name}.uri"
+    elsif method_obj['location'] == 'expression'
+      "#{method_obj.name}.expr"
```
gmcculloug: Is the expression data being saved in `YAML` format?  If so do we need a new file type or can we just use `.yaml`?

mkanoor: @gmcculloug We use the YAML extension to store method properties file, and we fetch all the methods based on the extension.

mkanoor:
The yaml file looks like

``` YAML
---
object_type: method
version: 1.0 
object:
  attributes:
    name: ExpTest
    display_name: 
    description: 
    scope: instance
    language: ruby
    location: expression
    data: |
      --- 
      :db: AvailabilityZone
      :expression:
        "=":
          field: AvailabilityZone-href_slug
          value: :user_input
  inputs:
  - field:
      aetype: 
      name: arg1
      display_name: 
      datatype: 
      priority: 1
      owner: 
      default_value: "${/#prefix}"
      substitute: false
      message: create
      visibility: 
      collect: 
      scope: 
      description: 
      condition: 
      on_entry: 
      on_exit: 
      on_error: 
      max_retries: 
      max_time: 
```

mkanoor:
Closed in favor of https://github.com/ManageIQ/manageiq-automation_engine/pull/56

PR #51, opened 2017-07-11
Move SyntaxChecker from manageiq-gems-pending

Fryguy:
The only consumer of this class is automate (specifically the MiqAeMethod model), however I feel that the engine should be the thing that provides the class, and the model should just use it, hence why I am moving here.


Fryguy:
This is more-or-less a straight move from manageiq-gems-pending, so I will not be fixing any rubocops here.

PR #53, opened 2017-07-19
[Depends on ManageIQ/manageiq#15615] Expose delete method for NetworkRouter

andyvesel:
This exposes the `delete` method for Network Router
Depends on https://github.com/ManageIQ/manageiq/pull/15615


gmcculloug:
@andyvesel Please review this and dependent PR to determine if they are still needed.

PR #55, opened 2017-07-26
Remove container specs

mkanoor:
Removed container_definition from service models and specs


gmcculloug:
This PR should resolve test failures caused by PR https://github.com/ManageIQ/manageiq/pull/15393.

PR #57, opened 2017-07-27
Add tests for CloudSubnet exposed methods

andyvesel:
Added tests for exposed `update` and `delete` methods for CloudSubnet


gmcculloug:
@andyvesel Can you review this PR and if still need please review failing tests.  Looks like `update_cloud_subnet` does not exist on the backing model.

PR #58, opened 2017-07-28
Expose operations for security group

andyvesel:
Exposed `update` and `delete` methods for SecurtyGroup
Depends on https://github.com/ManageIQ/manageiq/pull/15675


gmcculloug:
Closing now that ManageIQ/manageiq#15675 is closed.

PR #61, opened 2017-07-31
Need to pass the user's group in to automate when the provision starts.

lfu:
User's current group might have changed before the provision finishes.
So the user's current group from DB may be different from the user's group when the request is sent into automate.

Part of https://github.com/ManageIQ/manageiq/pull/15696.
~~Blocks https://github.com/ManageIQ/manageiq-content/pull/156.~~
Includes https://github.com/ManageIQ/manageiq/pull/15760.

https://bugzilla.redhat.com/show_bug.cgi?id=1467364

@miq-bot add_label bug, euwe/yes, fine/yes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -305,9 +305,11 @@ module MiqAeEngine
 
   def self.ae_user_object(options = {})
     raise "user_id not specified in Automation request" if options[:user_id].blank?
-    # raise "group_id not specified in Automation request" if options[:miq_group_id].blank?
+    # raise "miq_group_id not specified in Automation request" if options[:miq_group_id].blank?
+
     User.find_by!(:id => options[:user_id]).tap do |obj|
-      # obj.current_group = MiqGroup.find_by!(:id => options[:miq_group_id])
+      obj.current_group = MiqGroup.find_by!(:id => options[:miq_group_id]) unless options[:miq_group_id] == obj.current_group.id
+      $miq_ae_logger.info("User [#{obj.userid}] with current group ID [#{obj.current_group.id}] name [#{obj.current_group.description}]")
     end
   end
 end
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
@@ -250,7 +250,7 @@ module MiqAeEngine
 
     def user_info_attributes(user)
       {'user' => user, 'tenant' => user.current_tenant, 'miq_group' => user.current_group}.each do |k, v|
-        value = MiqAeObject.convert_value_based_on_datatype(v.id, v.class.name)
+        value = MiqAeMethodService::MiqAeServiceModelBase.wrap_results(v)
         @attributes[k] = value unless value.nil?
       end
     end
--- a/spec/miq_ae_engine_spec.rb
+++ b/spec/miq_ae_engine_spec.rb
@@ -131,6 +131,7 @@ describe MiqAeEngine do
           args[:instance_name]    = "DEFAULT"
           args[:fqclass_name] = "Factory/StateMachines/ServiceProvision_template"
           args[:user_id] = @user.id
+          args[:miq_group_id] = @user.current_group.id
           expect(MiqAeEngine).to receive(:create_automation_object).with("DEFAULT", attrs, :fqclass => "Factory/StateMachines/ServiceProvision_template").and_return('uri')
           expect(MiqAeEngine.deliver(args)).to eq(@ws)
         end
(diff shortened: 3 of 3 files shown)


gmcculloug:
@lfu Please review failing tests.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
```diff
@@ -250,7 +250,7 @@ def process_assertions(message)
 
     def user_info_attributes(user)
       {'user' => user, 'tenant' => user.current_tenant, 'miq_group' => user.current_group}.each do |k, v|
-        value = MiqAeObject.convert_value_based_on_datatype(v.id, v.class.name)
+        value = MiqAeMethodService::MiqAeServiceModelBase.wrap_results(v)
```
lfu: This solves the issue that the user object is got from DB with wrong current group in automate.

```MiqAeObject.convert_value_based_on_datatype``` calls ```find``` at service model layer which fetches the object from DB then calls ```MiqAeMethodService::MiqAeServiceModelBase.wrap_results``` on the object.

Since the user object is already available with the correct current group, we can call ```MiqAeMethodService::MiqAeServiceModelBase.wrap_results``` on it directly.

With this change, then https://github.com/ManageIQ/manageiq-content/pull/156 is not needed as the in-memory user object has the current group point to the group that the user is in when the request is made.

Comment (lfu) [inline lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb]: @mkanoor The ```find``` gets the user object from DB which is not what we want here since the user's current group from DB is not the correct requester group. 

We should use the in-memory user object that we already have at hand and wrap it into service model object. This way we don't need to change the expose of ```current_group``` in ```MiqAeServiceUser``` because the user object has got correct current group in memory. ```expose``` itself does not query the DB.

**Comment by lfu** [comment]
@miq-bot remove_label wip

**Comment by gmcculloug** [comment]
@mkanoor Please take another look.

File: lib/miq_automation_engine/engine/miq_ae_engine.rb [hunk 2]
```diff
@@ -305,9 +305,11 @@ def self.resolve_automation_object(uri, user_obj, attr = nil, options = {}, read
 
   def self.ae_user_object(options = {})
     raise "user_id not specified in Automation request" if options[:user_id].blank?
-    # raise "group_id not specified in Automation request" if options[:miq_group_id].blank?
+    raise "miq_group_id not specified in Automation request" if options[:miq_group_id].blank?
```
gmcculloug: @lfu Do you know why this was commented out before?  Are there any code paths you are aware of that might run into this new requirement?

lfu: @gmcculloug ```miq_group_id``` was not required because a user's ```current_group``` is used as default.
Seems ```AutomationTask``` is not passing ```miq_group_id``` when calling into automate: [here](https://github.com/ManageIQ/manageiq/blob/731174c7c4cb6f21a61c8f3270c5fca5d6513866/app/models/automation_task.rb#L25). Will fix it in another PR.

gmcculloug: Ok, please like the related PR back here when it is ready, I would like to have it available before merging this one since we know it will break Automation requests.

lfu: Add https://github.com/ManageIQ/manageiq/pull/15760.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -305,9 +305,11 @@ def self.resolve_automation_object(uri, user_obj, attr = nil, options = {}, read
 
   def self.ae_user_object(options = {})
     raise "user_id not specified in Automation request" if options[:user_id].blank?
-    # raise "group_id not specified in Automation request" if options[:miq_group_id].blank?
+    # raise "miq_group_id not specified in Automation request" if options[:miq_group_id].blank?
```
gmcculloug: The `raise` will be re-introduced upstream after this PR is merged.  This will allow the PR to be back-ported and to older branch and give more burn-in time upstream.

simaishi:
Fine backport (to manageiq repo) details:

```
$ git log -1
commit a3bb99ccd92d428c2a7c8218eb09b6e92c731500
Date:   Wed Aug 9 17:01:12 2017 -0400

    Merge pull request #61 from lfu/service_user_in_groups_1467364
    
    Need to pass the user's group in to automate when the provision starts.
    (cherry picked from commit 8de13ba4c7413710e4481cd7484b3675a12bd517)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1480007
```

simaishi:
Euwe backport (to manageiq repo) details:

```
$ git log -1
commit eea7b864c36f4cb1dadc87043bc2c8844c48c4db
Date:   Wed Aug 9 17:01:12 2017 -0400

    Merge pull request #61 from lfu/service_user_in_groups_1467364
    
    Need to pass the user's group in to automate when the provision starts.
    (cherry picked from commit 8de13ba4c7413710e4481cd7484b3675a12bd517)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1481859
```

PR #62, opened 2017-08-03
[WIP]Edit exposed methods for LoadBalancer

andyvesel:
Changed exposed `raw` methods to `non-raw`, according to pattern, where we use `non-raw` methods in `MiqAeService` modules
Depends on https://github.com/ManageIQ/manageiq/pull/15720


andyvesel:
@simaishi do we need a BZ for this?

simaishi:
@andyvesel If this is a change for 'master' branch only, it's not required to have a BZ. PRs that need to be backported will require a BZ unless it affects test only.

andyvesel:
@simaishi thanks, got it

andyvesel:
According to https://github.com/ManageIQ/manageiq/pull/15720#issuecomment-320034194 we may not want to do this

PR #67, opened 2017-09-01
Add exposed manager refresh method

Ladas:
Add exposed manager refresh method

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_ems_event.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_ems_event.rb
@@ -1,5 +1,9 @@
 module MiqAeMethodService
   class MiqAeServiceEmsEvent < MiqAeServiceEventStream
+    def manager_refresh(sync: false)
+      ar_method { @object.manager_refresh(:sync => sync) }
+    end
+
     def refresh(*targets, sync)
       ar_method { @object.refresh(*targets, sync) } unless targets.blank?
     end


lib/miq_automation_engine/service_models/miq_ae_service_ems_event.rb
```diff
@@ -1,5 +1,9 @@
 module MiqAeMethodService
   class MiqAeServiceEmsEvent < MiqAeServiceEventStream
+    def manager_refresh(sync: false)
```
gmcculloug: @Fryguy @jrafanie Have we started using Ruby keyword arguments in other places yet?  Is there any objection to this?

jrafanie: I don't know if we have but I don't see why not.  I know @chrisarcand would approve.  I think kwargs were feature complete by ruby 2.1 with required kwargs.

Ladas: :heart: kwargs :-) I use them in many places :-)

chrisarcand: Keyword arguments were added in 2.0 and you're more than welcome to use them. 👍

gmcculloug: I'm all for it, just wanted to check if there were any reasons we might be avoiding it.  I see now that it is being used by the backend method this service model is calling: https://github.com/ManageIQ/manageiq/blob/9291810e126cd34e5931bbf985be430a7018a422/app/models/ems_event/automate.rb#L5  
👍

PR #68, opened 2017-09-06
Expose stack refresh method

Ladas:
Expose stack refresh method

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
@@ -34,7 +34,6 @@ module MiqAeMethodService
 
     def self.refresh(manager_id, manager_ref)
       OrchestrationStack.refresh_ems(manager_id, manager_ref)
-      true
     end
   end
 end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
@@ -11,6 +11,7 @@ module MiqAeMethodService
     expose :raw_delete_stack
     expose :raw_update_stack
     expose :raw_exists?
+    expose :refresh, :method => :refresh_ems
 
     def add_to_service(service)
       error_msg = "service must be a MiqAeServiceService"
--- a/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
@@ -31,5 +31,10 @@ module MiqAeMethodService
     rescue MiqException::MiqOrchestrationStackNotExistError => err
       ['not_exist', err.message]
     end
+
+    def self.refresh(manager_id, manager_ref)
+      OrchestrationStack.refresh_ems(manager_id, manager_ref)
+      true
+    end
   end
 end
--- a/spec/service_models/miq_ae_service_orchestration_stack_spec.rb
+++ b/spec/service_models/miq_ae_service_orchestration_stack_spec.rb
@@ -32,5 +32,19 @@ module MiqAeServiceOrchestrationStackSpec
         expect(service_stack.normalized_live_status).to eq(['not_exist', 'test failure'])
       end
     end
+
+    context "refresh" do
+      before { stack.update(:ext_management_system => FactoryGirl.create(:ext_management_system)) }
+
+      it "calls a refresh on OrchestrationStack object" do
+        expect(stack.class).to receive(:refresh_ems).with(stack.ext_management_system.id, stack.ems_ref)
+        service_stack.refresh
+      end
+
+      it "calls a refresh on OrchestrationStack class" do
+        expect(stack.class).to receive(:refresh_ems).with(stack.ext_management_system.id, stack.ems_ref)
+        service_stack.class.refresh(stack.ext_management_system.id, stack.ems_ref)
+      end
+    end
   end
 end


lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
```diff
@@ -11,6 +11,7 @@ class MiqAeServiceOrchestrationStack < MiqAeServiceModelBase
     expose :raw_delete_stack
     expose :raw_update_stack
     expose :raw_exists?
+    expose :refresh, :method => :refresh_ems
```
mkanoor: @Ladas 
Can we call the method refresh_ems instead of refresh

```
expose :refresh_ems
```

Ladas: /we use the same naming in vm_or_template and other I think, so lets keep it consistent, or rename all of them?

Ladas: hm I think we can't rename them all really, we must be backwards compatible for Automate scripts, so vm.refresh must stay

Ladas:
@mkanoor specs added

Ladas:
ping @mkanoor @gmcculloug

lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
```diff
@@ -30,5 +31,10 @@ def normalized_live_status
     rescue MiqException::MiqOrchestrationStackNotExistError => err
       ['not_exist', err.message]
     end
+
+    def self.refresh(manager_id, manager_ref)
+      OrchestrationStack.refresh_ems(manager_id, manager_ref)
+      true
+    end
```
gmcculloug: @Ladas What is the return from `OrchestrationStack.refresh_ems`?  Why are you overriding it?

spec/service_models/miq_ae_service_orchestration_stack_spec.rb
```diff
@@ -32,5 +32,22 @@ module MiqAeServiceOrchestrationStackSpec
         expect(service_stack.normalized_live_status).to eq(['not_exist', 'test failure'])
       end
     end
+
+    context "refresh" do
+      before :each do
+        stack.ext_management_system = FactoryGirl.create(:ext_management_system)
+        stack.save
+      end
```
gmcculloug: `before(:each)` is the default, this can just be `before`.

Can you one-line this as:
```ruby
before { stack.update(:ext_management_system => FactoryGirl.create(:ext_management_system)) }
```

lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
Ladas: I think that just queues full/targeted refresh based on settings, so it should not have any return value

spec/service_models/miq_ae_service_orchestration_stack_spec.rb
Ladas: sure

lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
gmcculloug: Right, so why override the return value?

I would suggest just wrapping the results:

```ruby
def self.refresh(manager_id, manager_ref)
  wrap_results(OrchestrationStack.refresh_ems(manager_id, manager_ref))
end
```

@mkanoor thoughts?

Ladas: Honestly, I don't know, tell me what is correct. :-)

I've seens this ```return true``` in some places, as a safe way to avoid non mapped object returned, if we don't care about return value.

gmcculloug: It would be helpful if you could tell us what is returned from the method you are calling, then we could make an informed decision.  I think that's the piece that's missing here.

Ladas: https://github.com/Ladas/manageiq/blob/300deb4c9aa3f7d35401f4a61fee980921142490/app/models/orchestration_stack.rb#L147

The queue refresh is the last command

gmcculloug: Correct.

> what is returned from the method you are calling

It looks like `EmsRefresh.queue_refresh` returns either an array or IDs or `nil`.  (https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh.rb#L63) 

We do not have to convert that return value into a service model then just change the method to:
```ruby
def self.refresh(manager_id, manager_ref)
  OrchestrationStack.refresh_ems(manager_id, manager_ref)
end
```

mkanoor: @Ladas @gmcculloug 
if we can wrap the results of OrchestrationStack.refresh_ems(manager_id, manager_ref) we can just pass that thru. If its nil it should be ok.

Ladas: ok, removing the ```true```

Ladas:
@gmcculloug @mkanoor should be ready now

lib/miq_automation_engine/service_models/miq_ae_service_orchestration_stack.rb
```diff
@@ -30,5 +31,9 @@ def normalized_live_status
     rescue MiqException::MiqOrchestrationStackNotExistError => err
       ['not_exist', err.message]
     end
+
+    def self.refresh(manager_id, manager_ref)
```
gmcculloug: @Ladas Code looks fine, but one last question: What is the use case for having the class method for refresh from automate?

Ladas: Right, so the new targeted refresh is not queuing MIQ models. Because that requires that the OrchestrationStack record was already refreshed. 

So instead we have a generic ManagerRefresh::Target (https://github.com/Ladas/manageiq/blob/300deb4c9aa3f7d35401f4a61fee980921142490/app/models/orchestration_stack.rb#L163), that we can queue for the refresh. And we call it as a class method, since this doesn't have to be existing OrchestrationStack record in MIQ DB.

PR #71, opened 2017-09-20
Rename orchestration_template_vnfd

bzwei:
As described in https://github.com/ManageIQ/manageiq/issues/15985 all subclasses of `OrchestrationTemplate` are being moved to provider repos. So far the only found service model class is `OrchestrationTemplateVnfd`.

Do we see any other places in automate that need update?

@miq-bot add_label refactoring
@miq-bot assign @gmcculloug  
cc @mkanoor


gmcculloug:
I checked the content and engine repos and other references are to the base class.  This looks like it.

PR #78, opened 2017-09-26
Ansible Playbook Automate Method

mkanoor:
This PR adds the ability to launch an Ansible Playbook as a
 - State machine method, the method can run longer and we will internally do the retry
 - Regular Method, the playbook has to end execution in 10 minutes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -1,59 +1,112 @@
 module MiqAeEngine
   class MiqAePlaybookMethod
-    include ApplicationController::Filter::SubstMixin
+    PLAYBOOK_CLASS = ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptPayload
+
     def initialize(aem, obj, inputs)
       @workspace = obj.workspace
       @inputs    = inputs
       @aem       = aem
-      @aw = AutomateWorkspace.create(:input  => serialize_workspace(inputs),
-                                     :user   => @workspace.ae_user,
-                                     :tenant => @workspace.ae_user.current_tenant)
-      @runner_options = build_options_hash(aem, @aw.guid)
-    end
-
-    def manageiq_env
-      {
-        'api_token'  => Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api'),
-        'api_url'    => MiqRegion.my_region.remote_ws_url,
-        'guid'       => @aw.guid,
-        'MIQ_SCRIPT' => @contents,
-        'miq_group'  => @workspace.ae_user.current_group.description
-      }
     end
 
     def run
-      $miq_ae_logger.info("Playbook Method passing options to runner: #{@runner_options}")
-      begin
-        result = ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScriptPayload.run(@runner_options)
-      rescue => err
-        $miq_ae_logger.error("Playbook Method Ended with error #{err.message}")
-        raise MiqAeException::AbortInstantiation, err.message
+      if @workspace.persist_state_hash[method_key].present?
+        task_id = @workspace.persist_state_hash[method_key]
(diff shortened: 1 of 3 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -0,0 +1,112 @@
...
+      $miq_ae_logger.info("Playbook Method passing options to runner: #{playbook_options}")
+      begin
+        task_id = PLAYBOOK_CLASS.run(playbook_options)
+      rescue => err
+        $miq_ae_logger.error("Playbook Method Ended with error #{err.message}")
+        reset
+        raise MiqAeException::AbortInstantiation, err.message
+      end
+
+      @workspace.root['ae_state_started'].blank? ? wait_for_method(task_id) : check_task_status(task_id)
```
gmcculloug: Can you move `@workspace.root['ae_state_started'].blank?` into a method to make the intention is clear.

Something like
```ruby
def running_in_state_machine?
  @workspace.root['ae_state_started'].present?
end
```

Then this line is easier to read:
`running_in_state_machine? ? check_task_status(task_id) : wait_for_method(task_id)`

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -0,0 +1,112 @@
...
+        else
+          @workspace.root['ae_result'] = 'error'
+          reset
+          raise MiqAeException::Error, task.message
+        end
+      else
+        @workspace.root['ae_result'] = 'retry'
+        @workspace.root['ae_retry_interval'] = 1.minute
+        @workspace.persist_state_hash['automate_workspace_guid'] = @aw.guid
+        @workspace.persist_state_hash[method_key] = task_id
```
gmcculloug: Also for readability can we move this "retry" logic into a separate method.  If we end up supporting a way to set the interval this logic will likely become more complex and would be better in a separate method.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+    end
+
+    def manageiq_env
+      {
+        'api_token' => Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api'),
+        'api_url'   => MiqRegion.my_region.remote_ws_url,
+        'guid'      => @aw.guid,
+        'miq_group' => @workspace.ae_user.current_group.description
```
gmcculloug: Should we pas `user` like we do during services call?  https://github.com/ManageIQ/manageiq/blob/master/app/models/service_ansible_playbook.rb#L63-L72.  Or are you expecting the user to pull that object from the workspace if they need it?

Also, `guid` is not a clear name.  Possible alternatives: `ae_guid`, `ws_guid`, `workspace_guid`.

gmcculloug:
@mkanoor Please review the rubocop issues.  (GlobalVars warning for `$miq_ae_logger` can be ignored here.)

mkanoor:
@gmcculloug Please review

spec/miq_ae_playbook_method_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    let(:inputs) { { 'name' => 'Fred' } }
+
+    context "regular method" do
+      it "success" do
+        allow(described_class::PLAYBOOK_CLASS).to receive(:run).and_return(miq_task.id)
+        allow(MiqRegion).to receive(:my_region).and_return(FactoryGirl.create(:miq_region))
+        allow(AutomateWorkspace).to receive(:create).and_return(aw)
+        miq_task.update_status(MiqTask::STATE_FINISHED, MiqTask::STATUS_OK, "Done")
+        allow(MiqTask).to receive(:wait_for_taskid).and_return(miq_task)
+        allow(workspace).to receive(:update_workspace)
```
gmcculloug: @mkanoor This basic pattern repeat 6 times in these tests.  Can you DRY up this logic?

PR #83, opened 2017-10-09
Renamed vmdb_reference to href_slug

mkanoor:
Renamed the reference to vmdb objects in the AutomateWorkspace stored in the database to use href_slug:: instead of vmdb_reference::


mkanoor:
@syncrou Once this PR gets merged you might have to change your Ansible Playbook module to use href_slug:: instead of vmdb_reference::

PR #84, opened 2017-10-09
Changed deliver find_by to raise an error when the object is not found.

tinaafitz:
Several customer evm.logs show the following error:
MIQ(MiqQueue#deliver) Message id: [204945], Error: [undefined method `base_class' for NilClass:Class]
Which is caused by a Service request(and task) being deleted while Service provisioning is executing.

Changing the deliver find_by method will show that the ServiceTemplateProvisionTask does not exist.

https://bugzilla.redhat.com/show_bug.cgi?id=1467537

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -72,7 +72,7 @@ module MiqAeEngine
       automate_attrs = options[:attrs].dup
 
       if object_type
-        vmdb_object = object_type.constantize.find_by(:id => object_id)
+        vmdb_object = object_type.constantize.find_by!(:id => object_id)
         automate_attrs[create_automation_attribute_key(vmdb_object)] = object_id
         vmdb_object.before_ae_starts(options) if vmdb_object.respond_to?(:before_ae_starts)
       end


tinaafitz:
@mkanoor Please review.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -72,7 +72,7 @@ def self.deliver(*args)
       automate_attrs = options[:attrs].dup
 
       if object_type
-        vmdb_object = object_type.constantize.find_by(:id => object_id)
+        vmdb_object = object_type.constantize.find_by!(:id => object_id)
```
gmcculloug: Very happy to see this.  Can we get a test around this change?  Might want to consider break this if-block out into a smaller method.  The `deliver_queue ` method is scary big.  :scream:

gmcculloug:
@tinaafitz bump

mkanoor:
@tinaafitz 
https://github.com/ManageIQ/manageiq-automation_engine/blob/8b423de8e5caaed52c13560ce13e886b29c3ce8c/spec/miq_ae_engine_spec.rb#L803
This line is expecting find_by instead of find_by! so we have to change it

lib/miq_automation_engine/engine/miq_ae_engine.rb
gmcculloug: I have asked @pkomanek to look into refactoring this method so we can move forward with this change.

PR #85, opened 2017-10-10
Rename miq_password to password in AutomateWorkspace

mkanoor:
We already have a way of passing passwords into Automate using password::encrypted_value, for the AutomateWorkspace we were using miq_password::encrypted_value.

Changing it to be password to be consistent.


mkanoor:
@gmcculloug Please Review

PR #86, opened 2017-10-11
Get the current user if user context is set.

lfu:
PR https://github.com/ManageIQ/manageiq/pull/16089 allows the user_id/group_id/tenant_id to be recorded in the Queue and sets the user context when dispatching the item from the Queue.

This PR would use the user context if it is available when sending events to automate.

Blocks https://github.com/ManageIQ/manageiq/pull/16179.

https://bugzilla.redhat.com/show_bug.cgi?id=1487749
cc @mkanoor

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_event.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_event.rb
@@ -115,13 +115,16 @@ module MiqAeEvent
   end
 
   def self.call_automate(obj, attrs, instance_name, options = {})
-    user = obj.tenant_identity
+    user = User.current_user || obj.tenant_identity
     raise "A user is needed to raise #{instance_name} to automate. [#{obj.class.name}] id:[#{obj.id}]" unless user
 
     q_options = {
       :miq_callback => options[:miq_callback],
       :priority     => MiqQueue::HIGH_PRIORITY,
-      :task_id      => nil          # Clear task_id to allow running synchronously under current worker process
+      :user_id      => user.id,
+      :group_id     => user.current_group.id,
+      :tenant_id    => user.current_tenant.id,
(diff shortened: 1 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_event.rb
```diff
@@ -115,13 +115,16 @@ def self.process_result(ae_result, aevent)
   end
 
   def self.call_automate(obj, attrs, instance_name, options = {})
-    user = obj.tenant_identity
+    user = User.current_user || obj.tenant_identity
```
mkanoor: @lfu @gmcculloug 
When these events are raised and if the user is part of multiple groups which group should get used. Do we record the group with the events or vm ownership?

spec/miq_ae_event_spec.rb
```diff
@@ -124,6 +124,14 @@
 
           MiqAeEvent.raise_evm_event("vm_create", vm, :vm => vm)
         end
+
+        it "has current user" do
+          User.current_user = user
```
mkanoor: @lfu use User.with_user similar to https://github.com/ManageIQ/manageiq/blob/b978b61737bf8bffcca23ab3710500b1f677d4fe/spec/lib/rbac/filterer_spec.rb#L242

This we we dont have it set for the whole thread.

lib/miq_automation_engine/engine/miq_ae_event.rb
lfu: @mkanoor The user and the group is recorded in the Queue when the event is raised. Then the user is set to the group in the memory when the item get dequeued and processed.

lfu:
@mkanoor Please review.

PR #89, opened 2017-10-13
Ansible Playbook Method.

tinaafitz:
Changed to use playbook.run, added user, and group to manageiq_env hash, 
renamed miq_group to X_MIQ_Group, removed config_info hash key.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -32,7 +32,7 @@ module MiqAeEngine
                                      :user   => @workspace.ae_user,
                                      :tenant => @workspace.ae_user.current_tenant)
       playbook_options = build_options_hash
-      $miq_ae_logger.info("Playbook Method passing options: #{playbook_options}")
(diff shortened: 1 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -117,10 +123,10 @@ def current_info
     end
 
     def build_options_hash
-      config_info = YAML.load(@aem.data)
+      config_info = YAML.safe_load(@aem.data)
       config_info[:extra_vars] = MiqAeReference.encode(@inputs)
       config_info[:extra_vars][:manageiq] = manageiq_env
-      { :name => @aem.name, :config_info => config_info }
+      config_info
```
gmcculloug: Consider using `tap` here:
```ruby
def build_options_hash
  YAML.safe_load(@aem.data).tap do |config_info|
    config_info[:extra_vars] = MiqAeReference.encode(@inputs)
    config_info[:extra_vars][:manageiq] = manageiq_env
  end
end
```

tinaafitz:
@gmcculloug Made change.

PR #95, opened 2017-10-25
Added support for manageiq_connection

mkanoor:
The python client api expects the URL and token to be defined in the manageiq_connection hash.
When we pass extra_vars into Ansible playbook we need to set the
* url
* token

In the manageiq_connection hash.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -7,6 +7,8 @@ module MiqAeEngine
       @workspace = obj.workspace
       @inputs    = inputs
       @aem       = aem
+      @api_token = Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api')
+      @api_url   = MiqRegion.my_region.remote_ws_url
     end
 
     def run
@@ -95,12 +97,20 @@ module MiqAeEngine
 
     def manageiq_env
       {
-        'api_token'          => Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api'),
-        'api_url'            => MiqRegion.my_region.remote_ws_url,
+        'api_token'          => @api_token,
+        'api_url'            => @api_url,
         'user'               => @workspace.ae_user.href_slug,
         'group'              => @workspace.ae_user.current_group.href_slug,
         'automate_workspace' => @aw.href_slug,
         'X_MIQ_Group'        => @workspace.ae_user.current_group.description
+      }
+    end
+
+    def manageiq_connection_env
+      {
+        'token'       => @api_token,
+        'url'         => @api_url,
+        'X_MIQ_Group' => @workspace.ae_user.current_group.description
       }
     end
 
@@ -124,6 +134,7 @@ module MiqAeEngine
       @aem.options.tap do |config_info|
         config_info[:extra_vars] = MiqAeReference.encode(@inputs)
         config_info[:extra_vars][:manageiq] = manageiq_env
+        config_info[:extra_vars][:manageiq_connection] = manageiq_connection_env
       end
     end
   end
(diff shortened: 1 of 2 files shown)


mkanoor:
@syncrou @gmcculloug @tinaafitz 
Please review

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -7,6 +7,8 @@ def initialize(aem, obj, inputs)
       @workspace = obj.workspace
       @inputs    = inputs
       @aem       = aem
+      @api_token = Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api')
+      @api_url   = MiqRegion.my_region.remote_ws_url
```
gmcculloug: I would prefer these not be in the initialize method and instead use memoization methods:

```ruby
  def api_token
    @api_token ||= Api::UserTokenService.new.generate_token(@workspace.ae_user.userid, 'api')
  end

  def api_url
    @api_url ||= MiqRegion.my_region.remote_ws_url
  end
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -95,15 +97,23 @@ def mark_for_retry(task_id)
...
         'automate_workspace' => @aw.href_slug,
         'X_MIQ_Group'        => @workspace.ae_user.current_group.description
       }
     end
 
+    def manageiq_connection_env
+      {
+        'token'       => @api_token,
+        'url'         => @api_url,
+        'X_MIQ_Group' => @workspace.ae_user.current_group.description
```
syncrou: @mkanoor -We don't need 'X_MIQ_Group' here if it is already 'X_MIQ_Group' on line 105.  I only need it in `['manageiq_connection']` if the key is 'X-MIQ_Group' on line 105

mkanoor: @syncrou 
Its needed when making a connection since it goes into HTTP Header, The python api client team is not using it right now but they are going to need it. Since thats the only way you can change the users group at login time.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -95,15 +95,23 @@ def mark_for_retry(task_id)
...
         'automate_workspace' => @aw.href_slug,
         'X_MIQ_Group'        => @workspace.ae_user.current_group.description
       }
     end
 
+    def manageiq_connection_env
+      {
+        'token'       => api_token,
+        'url'         => api_url,
+        'X_MIQ_Group' => @workspace.ae_user.current_group.description
```
syncrou: Since it is being passed via extra_vars - e.g. `:extra_vars[:manageiq_connection]` the action plugin handles what gets passed down to the module, so the `X_MIQ_Group` key in both `:manageiq` and `:manageiq_connection` isn't needed. 

For consistency I would recommend just leaving it in `:extra_vars[:manageiq]`  Like you said, the ability to use 'X_MIQ_Group' to be used by the pyhon-api-client still needs to be added, pulling this option out of `:extra_vars[:manageiq_connection]` doesn't affect that change either way.

gmcculloug:
@mkanoor I assume we need this for gaprindashvili, correct?

mkanoor:
@gmcculloug This is needed for the G release

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
mkanoor: @syncrou 
Its needed for backward compatibility for playbooks that make direct REST API calls, we can remove all of them as a set in the next release

PR #98, opened 2017-10-30
Set state machine retry interval for Ansible playbook method.

tinaafitz:
Use execution_ttl to calculate the state machine retry_interval.
The Automate Method editor for Ansible playbooks accepts a value for the ttl(time to live) which is the expected time, in minutes, for a playbook to complete execution. Retries are only valid for state machines, so the Ansible Playbook method execution_ttl value is ignored when not executed within a state machine.  
The execution_ttl value is divided by the max_retries to determine the state machine retry interval.
For example, an execution_ttl of 700(minutes) in a state machine state that has a max_retries of 100, would result in a state machine retry interval of 7 minutes.
Any calculation resulting in a retry_interval less than 1 minute will be ignored and replaced by a 1 minute retry interval.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -2,6 +2,7 @@ module MiqAeEngine
   class MiqAePlaybookMethod
     PLAYBOOK_CLASS = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Playbook
     METHOD_KEY_SUFFIX = "_ansible_method_task_id".freeze
+    MIN_RETRY_INTERVAL = 1.minute
 
     def initialize(aem, obj, inputs)
       @workspace = obj.workspace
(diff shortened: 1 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -2,6 +2,7 @@ module MiqAeEngine
   class MiqAePlaybookMethod
     PLAYBOOK_CLASS = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Playbook
     METHOD_KEY_SUFFIX = "_ansible_method_task_id".freeze
+    MIN_RETRY_INTERVAL = 1
```
mkanoor: @tinaafitz 
Can we change this to 1.minute so we know the units.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -88,9 +89,23 @@ def check_task_status(task_id)
 
     def mark_for_retry(task_id)
       @workspace.root['ae_result'] = 'retry'
-      @workspace.root['ae_retry_interval'] = 1.minute
+      @workspace.root['ae_retry_interval'] = retry_interval.minutes
       @workspace.persist_state_hash['automate_workspace_guid'] = @aw.guid
       @workspace.persist_state_hash[method_key] = task_id
+      $miq_ae_logger.info("Setting retry interval: #{@workspace.root['ae_retry_interval']}")
```
mkanoor: @tinaafitz 
Should this message indicate that this is the internal retry value as opposed to the value set from the playbook.

mkanoor:
👍 Merge on green

PR #112, opened 2017-11-08
Add support for accessing objects from anywhere in the workspace

mkanoor:
https://bugzilla.redhat.com/show_bug.cgi?id=1350560

The Automation Workspace consists of resolved objects, which are instantiated from the database and optionally embellished by methods. These resolved objects contain the location from where they were instantiated which includes the domain, namespace(s), class and instance names.

e.g.
/ManageIQ/System/Request/MyInstance

Since each resolved object has this path, we can enhance our search to be based on the fully qualified path of the object. The current object search is limited to objects in the current leg of the resolution and doesn't allow access to earlier created objects in different parts of the resolution.

The enhanced search would help in 
 * Not using collect to migrate attributes between objects, since any object in the workspace is now accessible.
 * Substitution using objects from any where in the workspace

Additional Features
 * Object search can optionally include the domain name (**not recommended**)
 * Allows glob support for instance names only

Substitution examples
 * ${/System/Request/MyInstance#attr1}                     Specifies the namespace, class and instance  
 * ${/System/Request/*#attr1}                                       Uses any instance
 * ${/ManageIQ/System/Request/MyInstance#attr1}   uses the domain name - **not recommended**

Since a namespaces can contain other namespaces, its important to always specify the instance name. If you are sure that there is only one instance you can drop the instance name and use * instead.

Previously substitution limits the object search path to current leg of resolution

screen shot 2017-11-08 at 3 08 41 pm

Enhanced search path would allow us to search for objects anywhere in the workspace

screen shot 2017-11-08 at 3 10 36 pm

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
@@ -1,0 +1,56 @@
+module MiqAeEngine
+  module MiqAeObjectLookup
+    def initialize_obj_entries
+      @lookup_table = []
+    end
+
+    def add_obj_entry(fqns, klass, instance, object)
+      parts = fqns.split('/')
+      domain = parts.shift
+      namespace = parts.join('/')
+
+      @lookup_table << {:domain    => domain.downcase,
+                        :namespace => namespace.downcase,
+                        :klass     => klass.downcase,
+                        :instance  => instance.downcase,
+                        :object    => object}
+    end
+
+    def find_obj_entry(path)
+      query = path[0] == '/' ? parse_obj_path(path[1..-1]) : parse_obj_path(path)
+
+      entries = @lookup_table.select { |entry| entry[:klass] == query['class_name'] }
+      ientries = entries.select { |entry| instance_match?(query['instance'], entry) }
+      match = ientries.detect { |entry| fq_match?(query['domain'], query['partial_ns'], entry) }
+      match ||= ientries.detect { |entry| ns_match?(query['fqns'], entry) }
+      match[:object] if match
+    rescue MiqAeException::InvalidPathFormat => err
+      $miq_ae_logger.error(err.message)
+      nil
+    end
+
+    def ns_match?(ns, entry)
+      ns == entry[:namespace]
+    end
+
+    def fq_match?(domain, ns, entry)
+      domain == entry[:domain] && ns == entry[:namespace]
+    end
+
+    def instance_match?(instance, entry)
+      File.fnmatch(instance, entry[:instance], File::FNM_CASEFOLD)
+    end
+
+    def parse_obj_path(path)
+      parts = path.downcase.split('/')
+      raise MiqAeException::InvalidPathFormat, "#{path} is invalid. Need atleast Namespace/Class/Instance" if parts.length < 3
+      {
+        'instance'   => parts.pop,
+        'class_name' => parts.pop,
+        'fqns'       => parts.join('/'),
+        'domain'     => parts.shift,
+        'partial_ns' => parts.join('/')
+      }
+    end
+  end
+end
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
@@ -43,7 +43,7 @@ module MiqAeEngine
 
     def parse_obj_path(path)
       parts = path.downcase.split('/')
-      raise MiqAeException::InvalidPathFormat, "#{path} is invalid. Need atleast Namespace/Class/Instance" if parts.length < 3
+      raise MiqAeException::InvalidPathFormat, "#{path} is invalid. Need at least Namespace/Class/Instance" if parts.length < 3
       {
         'instance'   => parts.pop,
         'class_name' => parts.pop,
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
@@ -21,12 +21,16 @@ module MiqAeEngine
 
       entries = @lookup_table.select { |entry| entry[:klass] == query['class_name'] }
       ientries = entries.select { |entry| instance_match?(query['instance'], entry) }
-      match = ientries.detect { |entry| fq_match?(query['domain'], query['partial_ns'], entry) }
-      match ||= ientries.detect { |entry| ns_match?(query['fqns'], entry) }
-      match[:object] if match
+      find_best_match(ientries, query)
     rescue MiqAeException::InvalidPathFormat => err
       $miq_ae_logger.error(err.message)
       nil
+    end
+
+    def find_best_match(ientries, query)
+      match = ientries.detect { |entry| fq_match?(query['domain'], query['partial_ns'], entry) }
+      match ||= ientries.detect { |entry| ns_match?(query['fqns'], entry) }
+      match[:object] if match
     end
 
     def ns_match?(ns, entry)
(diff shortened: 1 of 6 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,56 @@
...
+      domain == entry[:domain] && ns == entry[:namespace]
+    end
+
+    def instance_match?(instance, entry)
+      File.fnmatch(instance, entry[:instance], File::FNM_CASEFOLD)
+    end
+
+    def parse_obj_path(path)
+      parts = path.downcase.split('/')
+      raise MiqAeException::InvalidPathFormat, "#{path} is invalid. Need atleast Namespace/Class/Instance" if parts.length < 3
```
tinaafitz: @mkanoor Missing a space between atleast.

tinaafitz:
@gmcculloug Please review.

mkanoor:
@gmcculloug 
Please review this PR is needed to get config information about email (from, signature, bcc etc)

gmcculloug:
> * Allows glob support for instance names only

Is this glob lookup case-insensitive?

> If you are sure that there is only one instance you can drop the instance name and use * instead.

What happens if there is more than one instance?  Is the selected instance nondeterministic or is an error raised?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,60 @@
...
+    def initialize_obj_entries
+      @lookup_table = []
+    end
+
+    def add_obj_entry(fqns, klass, instance, object)
+      parts = fqns.split('/')
+      domain = parts.shift
+      namespace = parts.join('/')
+
+      @lookup_table << {:domain    => domain.downcase,
```
mkanoor: @gmcculloug 
The automate model is case insensitive so we always lowercase the names when adding to the table.
But when we do the glob match we have to do a case-insensitive search
File.fnmatch(instance, entry[:instance], File::FNM_CASEFOLD)
Since the subst string is enter by the user they could use mixed case but we convert everything to lower case when doing a match. The fnmatch function needs the FNM_CASEFOLD to make it case insensitive.
If there are multiple instances we use the first one. If there is no match we return a nil and the subst call fails

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,60 @@
...
+      find_best_match(ientries, query)
+    rescue MiqAeException::InvalidPathFormat => err
+      $miq_ae_logger.error(err.message)
+      nil
+    end
+
+    def find_best_match(ientries, query)
+      match = ientries.detect { |entry| fq_match?(query['domain'], query['partial_ns'], entry) }
+      match ||= ientries.detect { |entry| ns_match?(query['fqns'], entry) }
+      match[:object] if match
```
mkanoor: @gmcculloug 
Return nil if we can't find a match. The detect gets the first match.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,60 @@
...
+                        :namespace => namespace.downcase,
+                        :klass     => klass.downcase,
+                        :instance  => instance.downcase,
+                        :object    => object}
+    end
+
+    def find_obj_entry(path)
+      query = path[0] == '/' ? parse_obj_path(path[1..-1]) : parse_obj_path(path)
+
+      entries = @lookup_table.select { |entry| entry[:klass] == query['class_name'] }
```
Fryguy: Since you are always looking up by entry[:klass], I think it would be better if the @lookup_table where a hash with the klass as the key (it can be part of the value too.  Then this line becomes an very efficient:

```ruby
@lookup_table[query['class_name']]
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,60 @@
...
+      ientries = entries.select { |entry| instance_match?(query['instance'], entry) }
```
Fryguy: What happens if the entries are not found?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,60 @@
...
+      find_best_match(ientries, query)
+    rescue MiqAeException::InvalidPathFormat => err
+      $miq_ae_logger.error(err.message)
+      nil
+    end
+
+    def find_best_match(ientries, query)
+      match = ientries.detect { |entry| fq_match?(query['domain'], query['partial_ns'], entry) }
+      match ||= ientries.detect { |entry| ns_match?(query['fqns'], entry) }
```
Fryguy: Again with all of these, you are ultimately doing a Array scanning operation.  It is preferable to construct the lookup table and the sub-tables so that you are doing optimal Hash lookups instead of Array scans.  The operations like fq_match? and ns_match? are very simplistic, so constructing the lookup table with keys like, for fq_match: `[domain, namespace]`

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
mkanoor: @Fryguy The substitution call handles it we just return a nil here.
https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb#L399

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
mkanoor: @Fryguy The class can be in multiple domains.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
mkanoor: @Fryguy Because of domain level overrides and partial matches (just namespaces) the search has to be different. The user can call substituion with fully qualified domain/namespace/class/instance or just namespace/class/instance.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
Fryguy: Still could use a lookup table, just store arrays by class

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
```diff
@@ -0,0 +1,59 @@
...
+    def initialize_obj_entries
+      @lookup_hash = Hash.new { |h, k| h[k] = [] }
+    end
+
+    def add_obj_entry(fqns, klass, instance, object)
+      parts = fqns.split('/')
+      domain = parts.shift
+      namespace = parts.join('/')
+      key = klass.downcase
+      @lookup_hash[key] = @lookup_hash[key] << {:domain    => domain.downcase,
```
Fryguy: I don't understand the assignment part of `@lookup_hash[key] = @lookup_hash[key] <<`  You can just modify it directly...

```ruby
@lookup_hash[key] << {:domain ...
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object_lookup.rb
Fryguy: The ientries is likely really small at this point because it's already gone through 2 filters, so a scan is probably ok

Fryguy:
Going to merge because the last comment is relatively minor.  @mkanoor Can you do a followup PR to clean that up?

PR #116, opened 2017-11-15
Allow automate scripts to use gems from GEM_PATH and BUNDLE_PATH

bdunne:
With bundler v1.16.0 the `BUNDLE_PATH` is not always on the `GEM_PATH`.  Ensure that gems from the `BUNDLE_PATH` are available to automate scripts by modifying the `ENV` before shelling out.

PR #119, opened 2017-11-21
Prefer Gem.path over ENV["GEM_PATH"]

bdunne:
Apparently the ENV variable isn't always there, but Gem.path is already an array of string paths.

Followup from PR https://github.com/ManageIQ/manageiq-automation_engine/pull/116

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -136,11 +136,10 @@ module MiqAeEngine
     private_class_method :run_ruby_method
 
     def self.with_automation_env
-      bundle_path = Bundler.bundle_path.to_s
+      gem_paths = (Gem.path + [Bundler.bundle_path.to_s]).uniq
       Bundler.with_clean_env do
         begin
           backup = ENV.to_hash
-          gem_paths = (backup["GEM_PATH"].split(File::PATH_SEPARATOR) << bundle_path).uniq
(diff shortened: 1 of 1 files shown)


bdunne:
@gmcculloug Please review.

gmcculloug:
@djberg96 Can you valid if this change fixes the issue for you?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -140,7 +140,7 @@ def self.with_automation_env
       Bundler.with_clean_env do
         begin
           backup = ENV.to_hash
-          gem_paths = (backup["GEM_PATH"].split(File::PATH_SEPARATOR) << bundle_path).uniq
+          gem_paths = (Gem.path << bundle_path).uniq
```
gmcculloug: @bdunne The `<<` operator is going to modify the `Gem.path` array, is that an intended action here?  Should this be something like `(Gem.path + [bundle_path])` instead?

djberg96:
👍

simaishi:

```
$ git log -1
commit 24e5b8e7425ac4dcc853c04d5f001c128664ce9a
Date:   Tue Nov 21 13:15:31 2017 -0500

    Merge pull request #119 from bdunne/change_gem_path_source
    
    Prefer Gem.path over ENV["GEM_PATH"]
    (cherry picked from commit 8b423de8e5caaed52c13560ce13e886b29c3ce8c)
```

Issue #107, opened 2017-11-06
Improve logging during copy operation to in case of failure


gmcculloug:
Followup work from https://bugzilla.redhat.com/show_bug.cgi?id=1461427

cc @d-m-u

JPrause:
@d-m-u with your reference to the PR above, does that mean this issue can be closed?
If yes, can you please close this issue, if not, can you remove the stale label

d-m-u:
Sure, can be closed, I would've, if I could close it. I have no power to do anything here.

JPrause:
Thanks @d-m-u 
@miq-bot close_issue

PR #120, opened 2017-11-21
Adds better logging to class copy methods

d-m-u:
Logs the variables we're using to save the class copy information for easier debugging when it goes kaput as part of https://github.com/ManageIQ/manageiq-automation_engine/issues/107

### DIFF
--- a/app/models/miq_ae_class_copy.rb
+++ b/app/models/miq_ae_class_copy.rb
@@ -79,13 +79,19 @@ class MiqAeClassCopy
 
   def validate
     dest_class = MiqAeClass.find_by_fqname("#{@target_ns_fqname}/#{@target_name}")
+    $log.info("Destination class: #{dest_class}")
     if dest_class
-      dest_class.destroy if @overwrite
+      $log.info("Overwrite flag: #{@overwrite}")
+      if @overwrite
+        dest_class.destroy
+        $log.info("Should only print if destination class exists and got destroyed by overwrite")
+      end
       raise "Destination Class already exists #{dest_class.fqname}" unless @overwrite
     end
   end
 
   def check_duplicity(domain, ns, classname)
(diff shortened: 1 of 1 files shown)


gmcculloug:
@mkanoor @tinaafitz This PR is intended to help if we run into the issue from https://bugzilla.redhat.com/show_bug.cgi?id=1461427 again since we were unable to reproduce.

app/models/miq_ae_class_copy.rb
```diff
@@ -79,13 +79,17 @@ def add_fields
 
   def validate
     dest_class = MiqAeClass.find_by_fqname("#{@target_ns_fqname}/#{@target_name}")
+    $miq_ae_logger.info("Destination class: #{dest_class}")
```
gmcculloug: @mkanoor Does the logging belong in the automation log or evm log or somewhere else?

app/models/miq_ae_class_copy.rb
```diff
@@ -79,13 +79,17 @@ def add_fields
...
     if dest_class
+      $miq_ae_logger.info("Overwrite flag: #{@overwrite}")
       dest_class.destroy if @overwrite
+      $miq_ae_logger.info("Should only print if destination class exists and got destroyed by overwrite")
```
tinaafitz: @d-m-u Should this log message be conditioned by "if @overwrite", since it states the destination class exists and got destroyed by overwrite, or maybe change the wording?

app/models/miq_ae_class_copy.rb
gmcculloug: @d-m-u Discussed with @mkanoor and agreed that this logging belongs in the evm.log so these calls should be using `$log`.

d-m-u:
@miq-bot bite me

PR #133, opened 2017-12-13
Fix issues in test

bdunne:
- Reference tested modules by their full name
- Don't wrap tests in a module


bdunne:
@gmcculloug Please review

PR #136, opened 2017-12-21
Support for internal methods

### LINKED ISSUE #135: introduce internal automate method
Currently every method in automate is configured to run in Drb and considered a user provided script. However manageiq is shipped with many automate methods which are not really user provided. We should introduce internal automate methods that run default implementations internally but allow users to override with an external script.

mkanoor:
Fixes #135

This enhancement allows you to call an internal method defined
in our app/models so we dont have to always run a method via DRb.
The default method implementation can be in app/models and if the
customer would like to overwrite the method they would add their
ruby method. The overhead for a customer script would be going
thru DRb.

The UI for an Internal Method would allow the user to configure the following options

 * target            :  The Service model object on which the method should be executed (allows for substitution) e.g. {/#miq_provision}

  * method         :  The name of the method
  * target_class  :  Service model class name, doesn’t allow for substitution (only needed if object is not specified)

  * output parameters : (optional)
      * result_obj    : The output of the method can be stored in an object in the workspace 
                valid values are
                 - **/** for the root object
                 - **.** for the current object
                 - **state_var** if the result should be stored as a state variable to persist between invocations.
      * result_attr         : the attribute name to store the result
      * retry_exception : true|false
      * retry_interval   : time interval in seconds

This is the existing interface for methods
Input_parameters
      User defined parameters are collected into a hash and passed into the method.
       Parameters can be substituted

Depends on ManageIQ PR https://github.com/ManageIQ/manageiq/pull/16715

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_internal_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_internal_method.rb
@@ -13,7 +13,12 @@ module MiqAeEngine
     def run
       validate_args
       obj = resolved_object || @aem.options[:class].constantize
-      run_method(obj, @aem.options[:method].to_sym)
+      method = @aem.options[:method].to_sym
+      if obj.respond_to?(method)
+        run_method(obj, method)
+      else
+        raise MiqAeException::MethodNotFound, "#{method} not defined for object #{obj}"
+      end
     end
 
     private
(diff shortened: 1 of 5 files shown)


mkanoor:
@bzwei Please review

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_internal_method.rb
```diff
@@ -0,0 +1,101 @@
...
+    def set_object_attribute(result, key)
+      value = if result.kind_of?(ActiveRecord::Base)
+                MiqAeMethodService::MiqAeServiceModelBase.wrap_results(result)
+              else
+                result
+              end
+      target_object[key || DEFAULT_ATTR_NAME] = value
+    end
+
+    def target_object
```
bzwei: Perhaps here we name it `result_object` or `resolve_result` if `target` is used for another meaning as I proposed earlier?

bzwei:
@mkanoor How can you make sure the object is a service model object? For the case of class method, can't the class be any valid class name? 
Why do you want to control how to process retry? `MiqAeRetryException` already means a retry. 
Is there a way for the object method to know whether it is the first time being executed or in the context of retry?

mkanoor:
@bzwei 
The target could point to an object any object in the workspace, all vmdb objects are wrapped in a service model by the engine so we dont have to check here. 
For Retry we need to figure out what the retry interval is, which can be configured by the user.

mkanoor:
Closing after discussion with Jason we should be using more builtin methods and encapsulate the object and workspace inside the builtin method.

PR #149, opened 2018-01-19
Builtin method for check_provisioned

mkanoor:
A builtin method is run directly in the Automate Engine and doesn't have to be run via DRb, so its faster.

This is an example of invoking the check_provisioned for VM and Instance provisioning.

Currently the check_provisioned method for vm and instance provisioning is an inline method
For testing you do would have to do the following steps

1. Create a new domain, mark it as enabled
2. Copy the following 2 methods to the new domain
 a. /ManageIQ/Infrastructure/VM/Provisioning/StateMachines/Methods/check_provisioned
 b. /ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods/check_provisioned
3. Copying the methods gives you the exact path of where the method should be with classes and namespaces
4. Delete the above 2 methods and add them back in as builtin
5. Make sure that the name is check_provisioned, builtin method names match to the internal name.

This method takes an optional single input parameter 
 

- ae_retry_interval                     Time in seconds to sleep between retries default is 60 seconds

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method/miq_ae_task_builtin.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method/miq_ae_task_builtin.rb
@@ -10,7 +10,7 @@ module MiqAeEngine
           raise ArgumentError, "miq_provision not specified" unless task
 
           result = task.statemachine_task_status
-          $miq_ae_logger.info("Builtin ProvisionCheck returned <#{result}> for state <#{task.state}> and status <#{task['status']}>")
+          $miq_ae_logger.info("Builtin check_provisioned returned <#{result}> for state <#{task.state}> and status <#{task['status']}>")
 
           set_ae_result(result, task, root, inputs)
         end
@@ -20,7 +20,7 @@ module MiqAeEngine
           when 'error'
             root['ae_result'] = 'error'
             reason = task.message
-            $miq_ae_logger.error("ProvisionCheck error <#{reason}>")
+            $miq_ae_logger.error("Builtin check_provisioned error <#{reason}>")
             reason = reason[7..-1] if reason[0..6] == 'Error: '
             root['ae_reason'] = reason
           when 'retry'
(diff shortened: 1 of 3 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method/miq_ae_task_builtin.rb
```diff
@@ -0,0 +1,36 @@
...
+      extend ActiveSupport::Concern
+
+      module ClassMethods
+        def miq_check_provisioned(obj, inputs = {})
+          root = obj.workspace.root
+          task = root['miq_provision']
+          raise ArgumentError, "miq_provision not specified" unless task
+
+          result = task.statemachine_task_status
+          $miq_ae_logger.info("Builtin ProvisionCheck returned <#{result}> for state <#{task.state}> and status <#{task['status']}>")
```
tinaafitz: @mkanoor Can we replace "ProvisionCheck" here with "check_provisioned"?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method/miq_ae_task_builtin.rb
```diff
@@ -0,0 +1,36 @@
...
+
+          set_ae_result(result, task, root, inputs)
+        end
+
+        def set_ae_result(result, task, root, inputs)
+          case result
+          when 'error'
+            root['ae_result'] = 'error'
+            reason = task.message
+            $miq_ae_logger.error("ProvisionCheck error <#{reason}>")
```
tinaafitz: @mkanoor Can we change it here as well?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method/miq_ae_task_builtin.rb
lfu: @mkanoor Are we going to have one file for all task built in methods?

mkanoor: @lfu I think we need 1 file for common functions like tasks (check_provisioned, check_service_provisioned, update_provision_status) They are all reading/updating the MiqRequestTasks

PR #151, opened 2018-01-19
Support for substitution from state_var

mkanoor:
The state_var stores key data that is needed between states, these
are typically user provided scalar data (ids, strings). The Automate
engine keeps the state_var segregated from the objects in the
workspace. The user might want to fetch one of these state var during
substitution. This PR allows access to the state vars using the
following syntax.

The format for substitution is

   ${object_path#attribute_name.optional_methods}
  The object path can be
   **.**                Default is missing stands for the current object
   **/**               The root object
   **/a/b/c**      The object in path /a/b/c

We are adding a new object path value called **STATE_VAR** (case insensitive)
to fetch attributes from state vars

e.g to get a state var called MY_TASK_ID use the following substitute string

**${STATE_VAR#MY_TASK_ID}**

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
@@ -396,7 +396,7 @@ module MiqAeEngine
 
     def fetch_object_attribute(path, name, required = false)
       o = @workspace.get_obj_from_path(path)
-      raise MiqAeException::ObjectNotFound, "Object Not Found for path=[#{path}]"  if o.nil?
+      raise MiqAeException::ObjectNotFound, "Object Not Found for path=[#{path}]" if o.nil?
 
       if o.attributes.key?(name.downcase)
         o.attributes[name.downcase]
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@gmcculloug Any chance to review this PR soon ?

For the V2V project, we have some states that rely on retry to wait for long tasks to finish. Unfortunately, $evm.root attributes don't survive retry and accessing state vars would be a good answer.

gmcculloug:
Overall the code changes and test look good.

My only concern is if `state_var` is the best keyword to use.  It makes sense because the users interact with it using `set_state_var` and `get_state_var` methods.  On the url we pass `ae_state_data` which is more of an internal value.  

Should this substitution value use a prefix like `ae_` to prevent a possible conflict?
Example: `(${ae_state_var#MY_TASK_ID}`
 
Any thoughts or suggestions on this?  @tinaafitz @pemcg

tinaafitz:
@gmcculloug I think `state_var` is safe here because it shouldn't show up as an object in that format.
It would have a `.` for the current object, `/` for the root object, or a `/a/b/c` path reference.

simaishi:

```
$ git log -1
commit 2dcc21ffc698e859d0195659bfce1ae0a350b8d9
Date:   Thu Feb 15 17:15:53 2018 -0500

    Merge pull request #151 from mkanoor/statevar_subst
    
    Support for substitution from state_var
    (cherry picked from commit 3d05a1b156bba9c893e08d3fc95b8a617c83f2b3)
```

PR #158, opened 2018-02-23
WIP: Support enumerable while loading objects

paulslaby:
This should be preparation for supporting `find_each` method in automate.
Maybe I have local testing suite broken, so I am trying to run the tests here. If it'll go ok, I'll fix mine.


paulslaby:
I understand. Do it'll be better to archive this PR.

PR #161, opened 2018-03-02
Fix removed have_attributes failures

chrisarcand:
Our custom have_attributes behavior of automatically comparing approximate times has been removed. You must specify this in the specs themselves now.

See https://github.com/ManageIQ/manageiq/pull/17021


chrisarcand:
@miq-bot add_labels test

@miq-bot assign @gmcculloug

PR #162, opened 2018-03-06
Change URI.split to replace double slashes with a single slash.

tinaafitz:
Replaced the double slashes with a single slash to be more forgiving in path substitution. 

Path example:  /{/#a}/{/#b}/{/#c}/{/#d}
before: if {/#a} is nil, path becomes //b/c/d which is invalid
after:  if {/#a} is nil, path becomes /b/c/d which is valid(*if b/c/d exists)

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1570131

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_uri.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_uri.rb
@@ -25,6 +25,7 @@ module MiqAeEngine
     end
 
     def self.split(uri, default_scheme = 'miqaews')
+      uri.gsub!(%r{/+}, '/')
       scheme, userinfo, host, port, registry, path, opaque, query, fragment = URI.split(uri.strip)
       scheme = default_scheme if scheme.nil?
       scheme = 'miqaews' if scheme.downcase == 'miqae'
--- a/spec/miq_ae_uri_spec.rb
+++ b/spec/miq_ae_uri_spec.rb
@@ -30,4 +30,12 @@ describe MiqAeEngine::MiqAeUri do
 
     expect(path).to eq(uri.strip)
   end
+
+  it "remove extra slashes" do
+    uri = "//Cloud//VM/StateMachines//Sample  "
+    result_uri = "/Cloud/VM/StateMachines/Sample"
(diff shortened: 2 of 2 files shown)


tinaafitz:
@miq-bot assign @gmcculloug
@mkanoor please review

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_uri.rb
```diff
@@ -26,6 +26,7 @@ def self.query2hash(query)
 
     def self.split(uri, default_scheme = 'miqaews')
       scheme, userinfo, host, port, registry, path, opaque, query, fragment = URI.split(uri.strip)
+      path = path.gsub('//', '/')
```
gmcculloug: Use `gsub!` so you do not need to reassign: `path.gsub!('//', '/')`

gmcculloug:
@tinaafitz Please review test failures.

d-m-u:
@tinaafitz is the [test failure](https://travis-ci.org/ManageIQ/manageiq-automation_engine/jobs/351376913#L2523) a concern?

d-m-u:
This spec is passing for me locally.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_uri.rb
```diff
@@ -26,6 +26,7 @@ def self.query2hash(query)
 
     def self.split(uri, default_scheme = 'miqaews')
       scheme, userinfo, host, port, registry, path, opaque, query, fragment = URI.split(uri.strip)
+      path&.gsub!('//', '/')
```
d-m-u: I think the line we want is more like ```uri.gsub! %r{/+}, '/'```

d-m-u:
screen shot 2018-04-20 at 8 31 53 am

tests green with line from comment

gmcculloug:
@tinaafitz The change feels more like a bug fix then an enhancement.  Do we think this needs to be back-ported?  If so we'll need a BZ to reference.

tinaafitz:
@gmcculloug I'm not sure this should be back-ported.
@mkanoor Thoughts?

PR #164, opened 2018-03-16
The factories are defined as symbols, update callers to use symbols

bdunne:
Fixes test failures after merging https://github.com/ManageIQ/manageiq/pull/16768


bdunne:

PR #175, opened 2018-04-25
Service model: method exposing

pkomanek:
* Adding the check for private attributes expose with spec. 
* Fixing a broken `base_class` and `base_model` methods in `miq_ae_service_model_base.rb` for classes with modules.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1560632

### DIFF
--- a/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
@@ -84,6 +84,13 @@ describe MiqAeMethodService::MiqAeServiceModelBase do
         expect(test_class.name).to eq('MiqAeMethodService::MiqAeServiceMiqAeServiceModelSpec_TestVmOrTemplate')
         expect(test_class.superclass.name).to eq('MiqAeMethodService::MiqAeServiceVmOrTemplate')
       end
+
+      it 'does not list private attributes' do
+        expect(MiqAeMethodService::MiqAeServiceModelBase).not_to(receive(:expose).with('properties'))
+        obj     = MiqAeServiceModelSpec::TestPrivateAttrExpose.create
+        svc_obj = described_class.model_name_from_active_record_model(MiqAeServiceModelSpec::TestPrivateAttrExpose).constantize.find(obj.id)
+        expect(svc_obj.methods.include?(:properties)).to(eq(false))
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -92,4 +98,12 @@ module MiqAeServiceModelSpec
   class TestInteger < ::Integer; end
   class TestApplicationRecord < ::ApplicationRecord; end
   class TestVmOrTemplate < ::VmOrTemplate; end
+  class TestPrivateAttrExpose < ::ApplicationRecord
+    self.table_name = 'generic_objects'
+
+    def properties
+      'private attr properties'
+    end
```
gmcculloug: I don't think we need to define this method since a getter method is exposed from ActiveRecord.  Just need to define the `private :properties` in the model.

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -68,6 +68,12 @@
       expect(described_class.create_service_model_from_name(:VmOrTemplate)).to be_nil
     end
 
+    it 'shouldn\'t expose private attr' do
+      allow(MiqAeMethodService).to receive(:const_get)
+      expect(MiqAeMethodService::MiqAeServiceModelBase).not_to receive(:expose).with('properties')
```
mkanoor: @pkomanek 
Can we add a spec to ensure that the methods function call doesn't list the method. like it is mentioned in the BZ.

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -84,6 +84,14 @@
         expect(test_class.name).to eq('MiqAeMethodService::MiqAeServiceMiqAeServiceModelSpec_TestVmOrTemplate')
         expect(test_class.superclass.name).to eq('MiqAeMethodService::MiqAeServiceVmOrTemplate')
       end
+
+      it 'does not list private attributes' do
+        expect(MiqAeMethodService::MiqAeServiceModelBase).not_to(receive(:expose).with('properties'))
+        obj     = MiqAeServiceModelSpec::TestPrivateAttrExpose.create
+        svc_obj = described_class.model_name_from_active_record_model(MiqAeServiceModelSpec::TestPrivateAttrExpose).constantize.find(obj.id)
+        expect(svc_obj.methods.include?(:properties)).to(eq(false))
+        obj.destroy
```
gmcculloug: Tests are run within database transactions, you do not need to cleanup the objects as part of the test.

PR #178, opened 2018-05-11
Ability to add and remove volumes for an instance 

djberg96:
Expose `create_volume_snapshot` and `cloud_volumes` so that users can add or remove a volume from an instance.

Modifications based on http://talk.manageiq.org/t/aws-cloud-volume-relationships/3111/2

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1507667

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_cloud_volume.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_cloud_volume.rb
@@ -9,6 +9,7 @@ module MiqAeMethodService
     expose :base_snapshot,          :association => true
     expose :cloud_volume_backups,   :association => true
     expose :cloud_volume_snapshots, :association => true
+    expose :create_volume_snapshot, :override_return => nil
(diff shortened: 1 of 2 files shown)


tinaafitz:
@djberg96 Looks good.

gmcculloug:
@djberg96 Please describe (or add link to) the required/optional parameters for the `create_volume_snapshot` method.

djberg96:
@gmcculloug Here's the links:

https://github.com/ManageIQ/manageiq-providers-amazon/blob/e3579a00d7dd04df3a275d71788cb6eeb715c499/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb#L86-L88

https://github.com/ManageIQ/manageiq-providers-amazon/blob/e3579a00d7dd04df3a275d71788cb6eeb715c499/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb#L26-L60

PR #179, opened 2018-05-14
Added method show_url for Service Model objects

mkanoor:
This allows automate methods to fetch the URL for request objects and
other objects with a simple call like

     svc_miq_request.show_url

instead of the current implementation


mkanoor:
Messed up the commits hence closing it

PR #180, opened 2018-05-14
Added method show_url for service model objects

mkanoor:
This allows automate methods to fetch the URL for request objects and
other objects with a simple call like

  svc_miq_request.show_url

Instead of

This is a common use case for sending emails to approvers with the
URL for the request to approve

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -8,6 +8,7 @@ module MiqAeMethodService
       include DRbUndumped  # Ensure that Automate Method can get at the class itself over DRb
     end
 
+    include Rails.application.routes.url_helpers
     include DRbUndumped    # Ensure that Automate Method can get at instances over DRb
     include MiqAeMethodService::MiqAeServiceObjectCommon
     include Vmdb::Logging
@@ -243,6 +244,13 @@ module MiqAeMethodService
       end
       @object = obj.kind_of?(ar_klass) ? obj : ar_method { ar_klass.find_by(:id => obj) }
       raise MiqAeException::ServiceNotFound, "#{ar_klass.name} Object [#{obj}] not found" if @object.nil?
+    end
+
+    def show_url
+      default_url_options[:host] = MiqRegion.my_region.remote_ui_url
+      url_for(:controller => ActiveModel::Naming.singular(self.class.ar_base_model),
+              :action     => 'show',
+              :id         => @object.id)
     end
 
     def virtual_columns_inspect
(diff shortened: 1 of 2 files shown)


mkanoor:
@gmcculloug @Fryguy 
Please review

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -245,6 +246,13 @@ def initialize(obj)
       raise MiqAeException::ServiceNotFound, "#{ar_klass.name} Object [#{obj}] not found" if @object.nil?
     end
 
+    def show_url
+      default_url_options[:host] = MiqRegion.my_region.remote_ui_url
```
gmcculloug: It would be great if there was a configuration setting that could specific a load balancer address and fall back to this address if it was empty.  cc @gtanzillo, @Fryguy

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -245,6 +246,13 @@ def initialize(obj)
...
+      url_for(:controller => ActiveModel::Naming.singular(self.class.ar_base_model),
+              :action     => 'show',
+              :id         => @object.id)
```
gmcculloug: What is the expected result if we do not have a controller for the resource being returned?  My guess is we are creating a URL that would look valid but not actually exposed.  For example, I do not think Lan/Switches have their own controller.  Also, since automate will dynamically create a default service model for any ActiveRecord backed object it could be almost anything.

Is there a way to return a `nil` value if we know it will be invalid?

mkanoor: @gmcculloug I don't think every vmdb object is going to be exposed, currently we have a need only for MiqRequest, VM and Service which the user can view externally. The decision to call this method is in the Automate methods, so either we can have a generic function that gets called selectively or have a constrained function that works on only a few vmdb objects, this PR   uses the former approach.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
mkanoor: @gmcculloug 
We allow for overrides of the web services url by reading from the settings.yml
https://github.com/ManageIQ/manageiq/blob/e915a3f7ef305c4a3d12cdb06b37e1d39902f6b2/app/models/miq_region.rb#L247
Maybe we can add the logic to do the same for remote_ui_server

himdel:
(I've already sent an email with this, but adding here for posterity :)...

I wonder whether it would make more sense to add a trampoline url,
something that would move the logic of `:controller =>
ActiveModel::Naming.singular(self.class.ar_base_model),` to the UI
instead of leaving it in Automate.

Such a trampoline already exists, created by @isimluk in https://github.com/ManageIQ/manageiq/pull/11607.

So, `/restful_redirect/?model=VmOrTemplate&id=123` already works :).
And you can add cases to `manageiq-ui-classic/app/controllers/restful_redirect_controller.rb`.

mkanoor:
Closed in favor of https://github.com/ManageIQ/manageiq-automation_engine/pull/181

isimluk:
Good think to know it still works @himdel!

Best wishes!

PR #181, opened 2018-05-16
Added show_url for specific service model objects

mkanoor:
Added a new public method show_url to the following service model objects
  - vm
  - service
  -  miq_request
  -  host
  -  ems_cluster
  -  storage
 
When we send links in emails we might need reference to the above objects

svc_miq_request.show_url

Instead of

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_ems_cluster.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_ems_cluster.rb
@@ -18,5 +18,9 @@ module MiqAeMethodService
       sync_or_async_ems_operation(false, "register_host", [host.id])
       true
     end
+
+    def show_url
+      MiqRegion.my_region.remote_ui_url + "/ems_cluster/show/" + "#{@object.id}"
+    end
   end
 end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_ems_cluster.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_ems_cluster.rb
@@ -20,7 +20,7 @@ module MiqAeMethodService
     end
 
     def show_url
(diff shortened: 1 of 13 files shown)


mkanoor:
@gmcculloug @Fryguy 
Please review

lib/miq_automation_engine/service_models/miq_ae_service_ems_cluster.rb
```diff
@@ -18,5 +18,9 @@ def register_host(host)
       sync_or_async_ems_operation(false, "register_host", [host.id])
       true
     end
+
+    def show_url
+      MiqRegion.my_region.remote_ui_url + "/ems_cluster/show/" + @object.id.to_s
```
Fryguy: Prefer here and everywhere else:

```ruby
URI.join(MiqRegion.my_region.remote_ui_url, "ems_cluster/show/#{@object.id}").to_s
```

spec/service_models/miq_ae_service_ems_cluster_spec.rb
```diff
@@ -0,0 +1,13 @@
+describe MiqAeMethodService::MiqAeServiceEmsCluster do
+  let(:cluster) { FactoryGirl.create(:ems_cluster) }
+  let(:svc_cluster) { MiqAeMethodService::MiqAeServiceEmsCluster.find(cluster.id) }
+
+  it "#show_url" do
+    ui_url = "https://www.example.com"
+    miq_region = FactoryGirl.create(:miq_region)
+    allow(MiqRegion).to receive(:my_region).and_return(miq_region)
+    allow(miq_region).to receive(:remote_ui_url).and_return(ui_url)
```
Fryguy: This setup is copy-pasted many times and doesn't use anything from the let blocks, so I think it's better to have this as a helper method in some spec/support file

mkanoor:
@Fryguy 
Please review

PR #183, opened 2018-05-21
Keep track of the server ids where the automate task has been processed.

lfu:
Part of https://github.com/ManageIQ/manageiq/pull/17451.

https://bugzilla.redhat.com/show_bug.cgi?id=1535237

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -65,6 +65,17 @@ module MiqAeEngine
     options[:ae_state_previous] = YAML.dump(ws.current_state_info) unless ws.current_state_info.empty?
   end
 
+  private_class_method def self.log_last_worker(worker_id)
+    msg = "Task was previously processed by Worker ID: [worker_id]"
+    w = MiqWorker.find_by(:id => worker_id)
+    if w.nil?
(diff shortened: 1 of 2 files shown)


lfu:
@miq-bot add_label 
gaprindashvili/no, enhancement

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -65,6 +65,13 @@ def self.deliver_queue(args, options = {})
     options[:ae_state_previous] = YAML.dump(ws.current_state_info) unless ws.current_state_info.empty?
   end
 
+  private_class_method def self.log_last_worker(worker_id)
+    w = MiqWorker.find_by(:id => worker_id)
```
mkanoor: @lfu Should we always log the worker id because its possible that the worker doesn't exist now, but was active a while back. When we have things stuck in the queue on a retry we would know which worker (which is expired now) was doing the work.

mkanoor: @lfu 
Maybe the log message should say that this worker doesn't exist anymore, and that could also help in debugging when an worker is constantly getting killed by some code that we might be running.

lfu:
@mkanoor Please review.

PR #188, opened 2018-06-15
Remove references to host provisioning

gmcculloug:
Remove files and specs related to Host Provisioning

Depends on https://github.com/ManageIQ/manageiq-content/pull/332


gmcculloug:
This feature was only supported on ESX hosts and from everything I know it was never used outside of some early testing.  I'm not expecting this to be an issue.

PR #191, opened 2018-06-25
Add vm.remove_disk method for a VMware VM

agrare:
Add a method to remove a disk by name to a VMware VM.

Depends on ManageIQ/manageiq-providers-vmware#290


agrare:
@miq-bot assign @gmcculloug

PR #206, opened 2018-08-06
Expose configuration_workflow and workflow_job

bzwei:
Add new service models in order to provision Ansible Tower workflow.

https://bugzilla.redhat.com/show_bug.cgi?id=1590975

Depends on ~~https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/103~~


bzwei:
Dependent PR merged.

PR #207, opened 2018-08-06
Allow for embedded methods to contain embedded methods

### LINKED ISSUE #202: [RFE] Multi-level dependency resolving with embedded method
Currently, adding an embedded method doesn't resolve its dependencies. For example, one needs to rename a virtual machine in a method. The implementation is different for RHV and VMware, so it would be easier to have a facade method that calls the actual method for the provider.

Say that the method that needs to rename the VM is `/Stuff/MyProcess/RenameVM`. To be able to call a method from the utility class, it embeds it, say `/Stuff/VM/Common/Rename`. The actual implementation is provided by `/Stuff/VM/VMware/Rename` and `/Stuff/VM/RHV`. Naively, one would simply embed these two methods in `/Stuff/VM/Common/Rename`. But recursive dependency is not implemented and the methods will have to be embedded in `/Stuff/MyProcess/RenameVM`.

This RFE asks for recursive dependency implementation, so that embedding methods is kept simple from the user stand point.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1609924

mkanoor:
Fixes #202

This PR allows for the embedded methods to embed other methods
It can handle if the methods are embedded multiple times
It can handle circular reference where any method can embed the top
level method.

Method A -> {Embeds Method B}
Method B -> {Embeds Method C and Method E}
Method C -> {Embeds Method D}
Method D -> {Embeds Method E}

_When the User uses Method A, we will automatically build a dependency list of (Method B, Method C, Method D and Method E). Even though Method E is embedded twice it will be only be listed once so the dependency list would be_

**Method B
Method C
Method D
Method E
Method A**

This whole payload is then fed to the ruby interpreter

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -286,7 +286,7 @@ module MiqAeEngine
     private_class_method :embedded_method_name
 
     def self.loaded?(current_items, fqname)
-      current_items.any? { |item| item[:fqname].casecmp(fqname) == 0 }
+      current_items.any? { |item| item[:fqname].casecmp(fqname).zero? }
     end
   end
 end
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -239,7 +239,9 @@ module MiqAeEngine
     private_class_method :cleanup
 
     def self.bodies_and_line_numbers(obj, aem)
-      embeds = embedded_methods(obj.workspace, aem) << {:data => aem.data, :fqname => aem.fqname}
+      embeds = []
+      embedded_methods(obj.workspace, aem, embeds, aem.fqname)
+      embeds << {:data => aem.data, :fqname => aem.fqname}
       code_start = 0
       script_info = {}
       bodies = []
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
Any chance to see it merged soon ? It would be useful for v2v (and others) because we use multi-level method embedding. That would make the method declaration cleaner by deduplicating embedded methods.

gmcculloug:
Initially we stated that we were avoiding this feature, and I believe the reasoning was the avoid complexity:
> No support for nested embedded methods

https://docs.google.com/presentation/d/1I8y8QCFDOjQtMRU6HUlCQWxe5sQm2Jm824dAM7SwthY/edit#slide=id.g270ae7d361_22_55

@mkanoor From the description you describe the load order.  I would have expected it to recursively load the methods by drilling down each path which would mean the load order would end up: **B, C, D, E, A**

Could you explain in a little more detail why the loading order is the way you described it.

@Fryguy Any concerns here?

Fryguy:
I don't understand why this is necessary... Duplicate definitions of a Ruby method work just fine

mkanoor:
@Fryguy 
This is necessary because an embedded method could have other methods embedded in it. Without this a user would have to list all the embedded methods at the top level. They would have to know if a method that they are embedding is embedding other methods and include it.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -274,5 +284,9 @@ def self.embedded_method_name(fqname)
       return parts.pop, parts.pop, parts.join('/')
     end
     private_class_method :embedded_method_name
+
+    def self.loaded?(current_items, fqname)
+      current_items.any? { |item| item[:fqname].casecmp(fqname).zero? }
```
Fryguy: @mkanoor this is what I was talking about... This is not necessary, imo, cause you can just allow the methods to double up

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -254,16 +256,24 @@ def self.bodies_and_line_numbers(obj, aem)
...
         match_ns = workspace.overlay_method(ns, klass, method_name)
         cls = ::MiqAeClass.find_by_fqname("#{match_ns}/#{klass}")
         aem = ::MiqAeMethod.find_by(:class_id => cls.id, :name => method_name) if cls
         raise MiqAeException::MethodNotFound, "Embedded method #{name} not found" unless aem
-        fqname = "#{match_ns}/#{klass}/#{method_name}"
-        $miq_ae_logger.info("Loading embedded method #{fqname}")
-        {:data => aem.data, :fqname => fqname}
+        fqname = "/#{match_ns}/#{klass}/#{method_name}"
+        if top == fqname
+          $miq_ae_logger.info("Skipping #{fqname}, cannot reference the top method")
```
Fryguy: Also don't understand the "top" distinction

mkanoor: @Fryguy thats to prevent circular referencing, a method can embed itself and we dont have any checks in the UI from stopping you doing that.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
mkanoor: @Fryguy I understand but we dont want to load the file multiple times and have ruby resolve it, if we can weed out the dupes before we have a final list it would be a better approach.

mkanoor: @Fryguy It could also lead to circular references so we want to stop that too

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
Fryguy: Ah ok...sounds like then there's other ways to get circulars as well... a includes b and b includes a for example

PR #208, opened 2018-08-07
Handle MIQ_STOP in state machine processing

mkanoor:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1441353

The MIQ_STOP exception was being suppressed in state machine
processing

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb
@@ -49,6 +49,10 @@ module MiqAeEngine
         $miq_ae_logger.warn("Skipping to state #{@workspace.root['ae_next_state']}")
         @workspace.root['ae_result'] = 'skip' if step == 'on_entry'
       end
+    rescue MiqAeException::StopInstantiation => e
+      error_message = "State=<#{f['name']}> running #{step} raised exception: <#{e.message}>"
+      $miq_ae_logger.error(error_message)
+      raise
     rescue => e
       error_message = "State=<#{f['name']}> running #{step} raised exception: <#{e.message}>"
       $miq_ae_logger.error(error_message)
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb
@@ -50,8 +50,7 @@ module MiqAeEngine
         @workspace.root['ae_result'] = 'skip' if step == 'on_entry'
       end
     rescue MiqAeException::StopInstantiation => e
-      error_message = "State=<#{f['name']}> running #{step} raised exception: <#{e.message}>"
-      $miq_ae_logger.error(error_message)
+      $miq_ae_logger.error("State=<#{f['name']}> running #{step} raised exception: <#{e.message}>")
       raise
     rescue => e
       error_message = "State=<#{f['name']}> running #{step} raised exception: <#{e.message}>"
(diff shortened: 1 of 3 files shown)


mkanoor:
@gmcculloug @tinaafitz 
Please review

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -165,7 +168,7 @@ def instantiate(uri, user, root = nil)
         raise
       ensure
         @current.pop if pushed
-        pop_state_machine_info if obj && obj.state_machine? && self.root
+        pop_state_machine_info if is_state_machine && self.root
```
gmcculloug: Can this just be `obj.try(:state_machine?)` ?  Then you do not need to defined the `is_state_machine` variable.

If it is not defined on the object it will return `nil` and you still have the desired outcome.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb
```diff
@@ -49,6 +49,10 @@ def process_state_step_with_error_handling(f, step = nil)
         $miq_ae_logger.warn("Skipping to state #{@workspace.root['ae_next_state']}")
         @workspace.root['ae_result'] = 'skip' if step == 'on_entry'
       end
+    rescue MiqAeException::StopInstantiation => e
+      error_message = "State=<#{f['name']}> running #{step} raised exception: <#{e.message}>"
+      $miq_ae_logger.error(error_message)
```
gmcculloug: Suggest moving the text interpolation directly into the `$miq_ae_logger.error` line as the `error_message` is only used once.

The rescue block below is setting a variable because it re-uses the results.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
gmcculloug: I know the code works as-is, but I'm not a fan of having the `is_state_machine` initially only set within 2 levels of if-blocks.  Would prefer setting `is_state_machine = false` after the `begin` statement and then only setting to `true` in the `if obj.state_machine?` check.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -136,6 +136,9 @@ def instantiate(uri, user, root = nil)
             save_current_state_info(@state_machine_objects.last) unless @state_machine_objects.empty?
             @state_machine_objects.push(obj.object_name)
             reset_state_info(obj.object_name)
+            is_state_machine = true
```
mkanoor: @gmcculloug 
These are the 2 places where the is_state_machine is set to either true or false

gmcculloug: Right, but that's not my point.  The variable is only set in a context that may never be called.  We only get into that block if `scheme == "miqaedb"`.  If `scheme` is set to something else we never initialize the variable, yet it is still referenced in the `ensure` block.

Maybe a better way to point this out is to look at the `push` variable that is also referenced from the `ensure` block.  It is initialized above the `begin` statement on line 131.  I think the `is_state_machine` variable should be initialized to `false` there.

d-m-u:
Other than GM's concern, looks great to me.

mkanoor:
@gmcculloug I have moved the is_state_machine to outside the block

billfitzgerald0120:
Results are different based on how the method is called.

I tested State MachineA calling State MachineB.  In one of the State MachineB steps,
a method with "exit MIQ_STOP" is called.

I am getting different results based on how the method is called. 
 
1. If you use an instance that connects to the method like : /Service/Provisioning/StateMachines/Methods/miq_stop

It will continue to the next step in State MachineB. Not expected

2. If you use METHOD::miq_stop syntax,

It will stop and return to State MachineA.  It will NOT run any more steps in State MachineB and return to the next step in State MachineA.   This is expected.

Billy

PR #209, opened 2018-08-08
Allow an Automate method to set a custom log prefix

mkanoor:
This PR allows customers method to set the log prefix.
Previously we have seen customers writing a custom log method to
add an additional string to the message.

2 new methods have been added

1. **$evm.set_log_prefix(string)**
2. **$evm.get_log_prefix**
 

e.g.

``` ruby
$evm.set_log_prefix("Custom Prefix")
$evm.log(:info, "Informational Message")
$evm.log(:error, "Error Message")
$evm.log(:warn, "Warning Message")
$evm.log(:info, "Current prefix is #{$evm.get_log_prefix}")
```

This would append the 'Custom Prefix' to every log info, warn, error and
debug

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
@@ -39,6 +39,7 @@ module MiqAeMethodService
       @workspace             = ws
       @persist_state_hash    = ws.persist_state_hash
       @logger                = logger
+      @log_prefix            = ""
       self.class.add(self)
       ws.disable_rbac
     end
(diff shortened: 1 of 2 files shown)


mkanoor:
@gmcculloug Please review

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -69,13 +70,25 @@ def disconnect_sql
...
 
     def set_state_var(name, value)
       @persist_state_hash[name] = value
     end
 
+    def set_log_prefix(prefix)
+      @log_prefix = prefix
+    end
+
+    def get_log_prefix
```
gmcculloug: Can we use `log_prefix=(prefix)` and `log_prefix` instead?

mkanoor: @gmcculloug 
Since this is for external consumption we have been using gets and sets so far.
$evm.get_state_var
$evm.set_state_var

So along the same lines I have kept it as
$evm.set_log_prefix &
$evm.get_log_prefix

gmcculloug:
I recall seeing some discussion with @pemcg on this feature.  It is not exactly clear to me how much it buys us moving this logic into the engine.  A user can simply set a global var in their script and reference it in the logging string.

Trying to figure out what the real benefit of adding this feature is.

pemcg:
So the real use-case is for customers that redirect their logs into something like Elasticsearch or Splunk, and so having a common prefix (or better a custom log format) can be really useful for searching. I'm assuming that defining the prefix at the $evm level affects all methods run in the context of that workflow, so there would much less to maintain than setting global vars in each method.

An even better solution would be the ability to define a system-wide prefix or better still logging format in system configuration, then it could be set once. That's why I was experimenting with custom logger classes as they seemed to be supported with `$evm.logger =`

PR #210, opened 2018-08-09
[WIP] Add submit_workflow option to provision request creation

d-m-u:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1613935

The submit_workflow option forces a specific path for how dialog field values are set up since everything gets funneled to the resource action workflow. Having this option in the create_service_provision_request allows the dialog options to be correctly passed to the service provisioning request. 

## Related to
https://github.com/ManageIQ/manageiq/pull/17839

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
@@ -111,7 +111,7 @@ module MiqAeMethodService
     end
 
     def self.create_service_provision_request(svc_template, options = nil)
-      result = svc_template.object_send(:provision_request, User.current_user, options)
+      result = svc_template.object_send(:provision_request, User.current_user, options, options[:submit_workflow] => true)
       MiqAeServiceModelBase.wrap_results(result)
     end
 


d-m-u:
@miq-bot add_label bug
@miq-bot assign @gmcculloug 
@miq-bot add_reviewer @eclarizio

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -111,7 +111,7 @@ def self.create_automation_request(options, userid = "admin", auto_approve = fal
     end
 
     def self.create_service_provision_request(svc_template, options = nil)
-      result = svc_template.object_send(:provision_request, User.current_user, options)
+      result = svc_template.object_send(:provision_request, User.current_user, options, options[:submit_workflow] => true)
```
mkanoor: @d-m-u @eclarizio 
Is the :submit_workflow a new option that was added, and does this need to be fixed even in the REST API?

mkanoor: I think the REST API might also need these changes.

https://github.com/ManageIQ/manageiq-api/blob/fa6ab6570125984c47f08990b3ce86877dee46b9/app/controllers/api/base_controller/generic.rb#L209

https://github.com/ManageIQ/manageiq-api/blob/fa6ab6570125984c47f08990b3ce86877dee46b9/app/controllers/api/services_controller.rb#L231

gmcculloug:
This PR does not fully resolves the BZ as it also requires the caller of `create_service_provision_request` to change from symbols to strings which should not be a required change.  We do it in our own automate method here: https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb#L21

We still need to identify the root cause of this change.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
gmcculloug: As @mkanoor points out this is change is required for more then just automate.  Can we push this change down to a common location like: https://github.com/ManageIQ/manageiq/blob/master/app/models/service_template.rb#L382

Suggesting this work be combined with https://github.com/ManageIQ/manageiq/pull/17839 as they are both trying to solve the same issue.

d-m-u:
Closing per https://github.com/ManageIQ/manageiq-automation_engine/pull/210#discussion_r209335806 cause this change should be in main repo.

PR #211, opened 2018-08-10
reload before #set_option and #get_option

bzwei:
fix https://bugzilla.redhat.com/show_bug.cgi?id=1614864

@miq-bot add_label bug
@miq-bot  assign @mkanoor 
cc @gmcculloug


lib/miq_automation_engine/service_models/mixins/miq_ae_service_miq_request_mixin.rb
```diff
@@ -4,10 +4,12 @@ def options
   end
 
   def get_option(key)
+    reload
```
gmcculloug: I am against adding the `reload` to the `get_*` methods, this could introduce a large performance overhead for minimal gain.  Users still have the option to call `reload` directly on the object if needed.

I would suggest holding off on this one as it is not needed to resolve this issue.

JPrause:
@miq-bot add_label blocker

JPrause:
@bzwei can you add the gaprindashvili/yes label if this can be backported.

bzwei:
We prefer another solution which is to add a new column just for cancellation. Closing this one.

bzwei:
Reopen to accept this a quick temporary solution.

lib/miq_automation_engine/service_models/mixins/miq_ae_service_miq_request_mixin.rb
```diff
@@ -17,6 +17,7 @@ def user_message=(msg)
 
   def set_option(key, value)
     ar_method do
+      reload
```
gmcculloug: @bzwei To protect this operation @Fryguy made the suggestion yesterday that we add a lock on the object here.

mkanoor: @Fryguy Should this method be in the ServiceModel or in the app/models? Maybe a new method that uses a transaction block. I am not sure if we have added locks or transaction blocks in the ServiceModel.

gmcculloug: My opinion is that this lock/reload/save block is only being added to address an automate method issue so there is no problem with the changes living here for now.

lib/miq_automation_engine/service_models/mixins/miq_ae_service_miq_request_mixin.rb
```diff
@@ -17,8 +17,10 @@ def user_message=(msg)
 
   def set_option(key, value)
     ar_method do
-      @object.options[key] = value
-      @object.update_attribute(:options, @object.options)
+      @object.with_lock do
+        @object.options[key] = value
+        @object.update_attribute(:options, @object.options)
```
gmcculloug: As rubocop suggest this should now be: `@object.update_attributes(:options => @object.options)`

lib/miq_automation_engine/service_models/mixins/miq_ae_service_miq_request_mixin.rb
```diff
@@ -17,8 +17,10 @@ def user_message=(msg)
 
   def set_option(key, value)
     ar_method do
-      @object.options[key] = value
-      @object.update_attribute(:options, @object.options)
+      @object.with_lock do
```
bzwei: We observed that `with_lock` automatically does a reload. It is confirmed in this discussion too https://stackoverflow.com/questions/15694326/rails-activerecord-pessimistic-locking-do-i-need-to-reload-after-obtaining-a-l

gmcculloug:
@Fryguy Any final thoughts here before merging?

simaishi:

```
$ git log -1
commit b17a6e22c7514a9281c98ffc370104c5c461bd37
Date:   Tue Aug 14 17:21:19 2018 -0400

    Merge pull request #211 from bzwei/reload_options
    
    Lock the model object when modify an option key
    (cherry picked from commit b601e3b32535c5a5a8201021c4f3d295acf39eaa)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1616076
```

PR #212, opened 2018-08-14
Enable cancelation_status to be updated for service model

hsong-rh:
Allow `cancelation_status` to be updated for service model.

https://bugzilla.redhat.com/show_bug.cgi?id=1614864

Related PRs:
https://github.com/ManageIQ/manageiq/pull/17825
https://github.com/ManageIQ/manageiq-schema/pull/254

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
@@ -4,6 +4,8 @@ module MiqAeMethodService
     expose :pre_ansible_playbook_service_template
     expose :post_ansible_playbook_service_template
     expose :mark_vm_migrated
+    expose :canceling
+    expose :canceled
 
     def transformation_destination(source_obj)
       ar_method do


lib/miq_automation_engine/service_models/miq_ae_service_miq_request_task.rb
```diff
@@ -20,6 +20,14 @@ def finished(msg)
       object_send(:update_and_notify_parent, :state => 'finished', :message => msg)
     end
 
+    def set_canceling
+      object_send(:update_attributes, :cancelation_status => 'canceling')
```
gmcculloug: @hsong-rh @bzwei Is this method still needed and if so it should be changed to call the backend `cancel` method.

https://github.com/ManageIQ/manageiq/pull/17825/files#diff-eb7db57808b5f20ff8f0f07bfd565890L105
But something still has to change, either the cancel method needs to be defined at the root level, maybe raising NotImplemented or this definition needs to move to the service_template_transformation_plan_task service model.

Also, I would expect this to just use the expose syntax above:
`expose :cancel`

lib/miq_automation_engine/service_models/miq_ae_service_miq_request_task.rb
```diff
@@ -20,6 +20,14 @@ def finished(msg)
...
+    end
+
+    def set_canceled
+      object_send(:update_attributes, :cancelation_status => 'canceled')
```
gmcculloug: Is this needed due to how v2v is controlling the cancel process?  If so this logic should still be code at the active_record layer and you just need to `expose` the method here.

gmcculloug:
@hsong-rh Did you try rebasing this PR as we discussed to resolve the Hakiri error?

hsong-rh:
@gmcculloug Now rebased.

PR #220, opened 2018-09-06
Fix Embedded Ansible playbook method retries in an automate state machine

mattparko:
This PR addresses https://bugzilla.redhat.com/show_bug.cgi?id=1625047

I have successfully tested the following scenarios:
* result=ok on first run
* X number of retries followed by result=ok
* max retries reached resulting in error
* playbook faliure resulting in error

The PR also fixes a retry count issue caused by entering the method twice in order to set status to 'async_launch'

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -15,10 +15,8 @@ module MiqAeEngine
       if @workspace.persist_state_hash[method_key].present?
         task_id = @workspace.persist_state_hash[method_key]
         @aw = AutomateWorkspace.find_by(:guid => @workspace.persist_state_hash['automate_workspace_guid'])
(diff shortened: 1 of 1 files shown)


mkanoor:
@mattparko The bug is here
https://github.com/ManageIQ/manageiq-automation_engine/blob/ee11ae28ad17e6becdde4a06316a1d749d9e5c0d/lib/miq_automation_engine/engine/miq_ae_engine.rb#L64
We clear out the state data from the playbook method but the engine reverts to the old value.
The fix is to add 2 lines to clear out the
ae_state_data and ae_state_previous 
I will be creating a PR shortly.
The other change you have is handled by the engine here
https://github.com/ManageIQ/manageiq-automation_engine/blob/ee11ae28ad17e6becdde4a06316a1d749d9e5c0d/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb#L24

gmcculloug:
Closing in favor of https://github.com/ManageIQ/manageiq-automation_engine/pull/222

PR #223, opened 2018-09-11
[WIP] Fixing ae_method calls for miq_ae_service_generic_object

pkomanek:
Purpose
---------------------
Fix for calling the user specified methods on service models for `generic_object`(GO) with the name from native built-in `ative_record` methods. We are changing the method call priority and making the GO to call custom methods at the first place and then native built-in methods. This is provided by calling the new method `go_send` from GO model inside the `method_missing` method.

This PR is based on https://bugzilla.redhat.com/show_bug.cgi?id=1607940 and it also depends on https://github.com/ManageIQ/manageiq/pull/18149.

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
@@ -28,8 +28,7 @@ module MiqAeMethodService
     def method_missing(method_name, *args)
       ae_user_identity unless @ae_user
       args = convert_params_to_ar_model(args)
-      results = object_send(method_name, *args)
-      wrap_results(results)
+      wrap_results(@object.go_send(method_name, *args))
     end
 
     def convert_params_to_ar_model(args)


lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
```diff
@@ -4,6 +4,14 @@ class MiqAeServiceGenericObject < MiqAeServiceModelBase
     include MiqAeServiceRemoveFromVmdb
     require 'drb'
 
+    def initialize(*args)
+      super(*args)
+
+      @object.property_methods.each do |method_name|
+        define_ae_method_from_generic_object(method_name)
+      end
```
gmcculloug: All this logic should be pushed into a method.  The method may loop over the method names and call another method to define the actual methods for this object, but I would expect a single, clear method name here.

Also, what about the `property_attributes`?  They need to be exposed as well.

lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
```diff
@@ -25,10 +33,23 @@ def ae_user_identity
...
+          wrap_results(@object.send(:_call_automate, method_name, *args))
+        end
+      end
+    end
+
     def method_missing(method_name, *args)
       ae_user_identity unless @ae_user
       args = convert_params_to_ar_model(args)
-      results = object_send(method_name, *args)
+      if @object.property_method_defined?(method_name)
```
gmcculloug: We did not want to keep the `method_missing` or `respond_to_missing?` methods in this file at all.  All Generic Object attributes and methods should be created in the `initialize` method which removes the need for this logic.

Also, the `ae_user_identity` and `convert_params_to_ar_model` logic would move to the dynamic method definition.

lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
```diff
@@ -28,7 +28,11 @@ def ae_user_identity
     def method_missing(method_name, *args)
       ae_user_identity unless @ae_user
       args = convert_params_to_ar_model(args)
-      results = object_send(method_name, *args)
+      results = if @object.property_method_defined?(method_name)
+                  @object.send(:_call_automate, method_name.to_s, *args)
+                else
+                  object_send(method_name, *args)
+                end
```
gmcculloug: While I like this I suggest that we move the logic to the ActiveRecord side to keep the internal GO logic there.  Maybe create a `go_send` or `generic_object_send` method and move this logic there.  We should try to keep the logic in the service model to the minimal we need to call AR object.

bzwei: Since generic_object is mainly used by automate, I would suggest the method in AR side to include not only `property_method`, but also `property_attribute` and `property_association`

PR #222, opened 2018-09-06
Clear stale data from ae_state_data and ae_state_previous

mkanoor:
https://bugzilla.redhat.com/show_bug.cgi?id=1625047

The ansible playbook was one of the first times we have cared
about what is in the ae_state_data, the playbook method clears out
the key ae_state_data fields regarding the automate_workspace and
the taskid that was used to track the playbook on a completion of the
playbook (either with 'ok' or 'retry'). Since the stale state data
wasn't being cleared the playbook wasn't getting successfully launched
on a retry and would bail looking for stale automate workspaces.

PR #224, opened 2018-09-11
[GAPRINDASHVILI] Clear stale data from ae_state_data and ae_state_previous

mkanoor:
The ansible playbook was one of the first times we have cared
about what is in the ae_state_data, the playbook method clears out
the key ae_state_data fields regarding the automate_workspace and
the taskid that was used to track the playbook on a completion of the
playbook (either with 'ok' or 'retry'). Since the stale state data
wasn't being cleared the playbook wasn't getting successfully launched
on a retry and would bail looking for stale automate workspaces.

Original PR https://github.com/ManageIQ/manageiq-automation_engine/pull/222

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1628657


gmcculloug:
@mkanoor Please add a link in the description to the original PR.

PR #228, opened 2018-09-13
Clean up the password field and value in automate and evm.log

lfu:
Hide password values in automate.log and evm.log.
Add "_id" attribute only for VMDB objects.

Depends on https://github.com/ManageIQ/manageiq-gems-pending/pull/373.
Includes https://github.com/ManageIQ/manageiq/pull/17986.

https://bugzilla.redhat.com/show_bug.cgi?id=1619385

@miq-bot add_label bug, gaprindashvili/yes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -8,9 +8,9 @@ module MiqAeEngine
   DEFAULT_ATTRIBUTES = %w( User::user MiqServer::miq_server object_name )
 
   def self.instantiate(uri, user)
-    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{uri}")
+    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{MiqPassword.sanitize_string(uri)}")
     workspace, t = Benchmark.realtime_block(:total_time) { MiqAeWorkspaceRuntime.instantiate(uri, user) }
-    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{uri}...Complete - Counts: #{format_benchmark_counts(t)}, Timings: #{format_benchmark_times(t)}")
+    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{MiqPassword.sanitize_string(uri)}...Complete - Counts: #{format_benchmark_counts(t)}, Timings: #{format_benchmark_times(t)}")
     workspace
   end
 
@@ -78,7 +78,7 @@ module MiqAeEngine
     begin
       miq_task&.state_active
       object_name = "#{options[:object_type]}.#{options[:object_id]}"
-      _log.info("Delivering #{options[:attrs].inspect} for object [#{object_name}] with state [#{state}] to Automate")
+      _log.info("Delivering #{MiqPassword.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate")
       automate_attrs = automate_attrs_from_options(options)
 
       if options[:object_type]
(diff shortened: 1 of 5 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
```diff
@@ -854,5 +854,11 @@ def classify_value(value)
         return self[value]
       end
     end
+
+    def attribute_for_vmdb_object?(klass, value)
+      Kernel.const_defined?(klass) && value.to_i.nonzero?
+    rescue NameError
```
gmcculloug: Should  use `klass.safe_constantize` here which will return `nil` if the class does not exist.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -310,4 +310,10 @@ def self.ae_user_object(options = {})
       $miq_ae_logger.info("User [#{obj.userid}] with current group ID [#{obj.current_group.id}] name [#{obj.current_group.description}]")
     end
   end
+
+  V2_PASSWORD_REGEXP = /v([0-9]+)%3A%7B(.*?)%7D/
```
gmcculloug: Push this to the top of the file to group with the other constant.  Also, I would recommend adding a comment clarifying why this one is different then the one in `MiqPassword`.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -310,4 +310,10 @@ def self.ae_user_object(options = {})
...
+
+  def self.sanitize_string(str)
```
gmcculloug: I would recommend changing this name to be more specific since it is really handling the special use-case of an encoded string.

Maybe: `sanitize_encoded_string`

@bdunne Do you think this should live in MiqPassword?  A bit of a special case.

JPrause:
@miq-bot add_label blocker

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -6,11 +6,12 @@
 
 module MiqAeEngine
   DEFAULT_ATTRIBUTES = %w( User::user MiqServer::miq_server object_name )
+  V2_PASSWORD_REGEXP = /v([0-9]+)%3A%7B(.*?)%7D/ # for URI encoded string "v2:{.*}"
 
   def self.instantiate(uri, user)
-    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{uri}")
+    $miq_ae_logger.info("MiqAeEngine: Instantiating Workspace for URI=#{sanitize_encoded_string(uri)}")
```
mkanoor: @lfu Why can't we use MiqPassword.sanitize_string here?

gmcculloug: The string is URL encoded so it no longer matches the regex in MiqPassword.

mkanoor: @lfu @gmcculloug 
Are we getting passwords as URL's?
Can the function sanitize_encoded_string to MiqPassword so its all in one place and we can control if it should be asterisk or [FILTERED] or something else

lfu:
```[FILTERED]``` is used by [log_hashes](https://github.com/ManageIQ/manageiq-gems-pending/blob/5ad6432693711f34c06246248ca892b8e58500dd/lib/gems/pending/util/vmdb-logger.rb#L120). ```********``` is used by ```MiqPassword.sanitize_string```.

File: lib/miq_automation_engine/engine/miq_ae_engine.rb (hunk 4 above)
Comment (mkanoor) [reply lib/miq_automation_engine/engine/miq_ae_engine.rb]: Its the password coming in from Dialogs any dialog_field that is marked as encrypted comes in with the v2 string.

**Comment by gmcculloug** [comment]
The `[FILTERED]` string also comes from rails https://guides.rubyonrails.org/action_controller_overview.html#parameters-filtering.

I would suggest we look into including the support for encoded password strings into MiqPassword as @mkanoor and I mentioned (https://github.com/ManageIQ/manageiq-automation_engine/pull/228#discussion_r217732856 and https://github.com/ManageIQ/manageiq-automation_engine/pull/228#discussion_r217502399) and the discussion about what replacement string to use should be a separate PR/issue.

**Comment by simaishi** [comment]

```
$ git log -1
commit 2452d0e82324067fb18782805ff5ad2872744fdf
Author: Madhu Kanoor <mkanoor@redhat.com>
Date:   Mon Sep 24 16:44:14 2018 -0400

    Merge pull request #228 from lfu/password_log_1619385
    
    Clean up the password field and value in automate and evm.log
    
    (cherry picked from commit f294b5636db07acfea58391230a7e8e41be73b1b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1634808
```

PR #231, opened 2018-09-18
I'm not sure why any of these tests are wrapped in modules

bdunne:


bdunne:
The rubocop comments are all due to the whitespace changes, so I'm not going to fix any of that here.

bdunne:
@gmcculloug or @mkanoor Please review

mkanoor:
@bdunne @gmcculloug Some of these specs are the old legacy ones that we inherited from the XML models without a domain.

PR #234, opened 2018-09-21
Patch DRb to prevent protocols other than UNIX

agrare:
To prevent the DRbTCPSocket protocol from being used patch
DRb::DRbProtocol to change the default to DRbUNIXSocket and prevent
add_protocol from being able to add other protocols.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
@@ -1,0 +1,19 @@
+require "drb/drb"
+require "drb/unix"
+
+# We are only interested in using UNIX Sockets, but the "feature" of looping
+# through all available protocols leads to issues where the default, TCP socket
+# is tried when a UNIX socket should be used.
+#
+# To prevent this we override the default protocol from TCP to UNIX and
+# prevent the addition of new protocols by overriding add_protocol.
+module DRb
+  module DRbProtocol
+    @protocol = [DRb::DRbUNIXSocket]
+
+    def add_protocol(_prot)
+      @protocol
+    end
+    module_function :add_protocol
+  end
+end
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -36,6 +36,8 @@ module MiqAeEngine
 
     def setup
       require 'drb/timeridconv'
+      require_relative 'miq_ae_drb_patches'
+
       global_id_conv = DRb::TimerIdConv.new(drb_cache_timeout)
       drb_front = MiqAeMethodService::MiqAeServiceFront.new(@workspace)
 


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
```diff
@@ -0,0 +1,17 @@
+require "drb/drb"
+require "drb/unix"
+
+# We are only interested in using UNIX Sockets, but the "feature" of looping
+# through all available protocols leads to issues where a TCP socket is
+# tried to be used when a UNIX socket should be used.
```
jrafanie: `where the default, TCP socket, is tried when a UNIX socket should be used.`

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# To prevent this we override the default protocol from TCP to UNIX and
+# prevent the addition of new protocols by overriding add_protocol
```
jrafanie: looks like a sentence so, missing a period `.` at the end

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
```diff
@@ -0,0 +1,17 @@
...
+module DRb::DRbProtocol
```
jrafanie: Can we use expanded notation?  We've already achieved extreme hack territory, I don't think we need constant problems too.  😉

agrare: Just setting you up for a future fix and/or bad joke ;)

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_drb_patches.rb
```diff
@@ -0,0 +1,19 @@
...
+# through all available protocols leads to issues where the default, TCP socket
+# is tried when a UNIX socket should be used.
+#
+# To prevent this we override the default protocol from TCP to UNIX and
+# prevent the addition of new protocols by overriding add_protocol.
+module DRb
+  module DRbProtocol
+    @protocol = [DRb::DRbUNIXSocket]
+
+    def add_protocol
```
Fryguy: Does add_protocol take parameters?

agrare: 🤦 yes it does

agrare:
Before requiring:
```
DRb::DRbProtocol.instance_variable_get(:@protocol)
[DRb::DRbTCPSocket]
DRb::DRbProtocol.add_protocol(DRb::DRbTCPSocket)
[DRb::DRbTCPSocket, DRb::DRbTCPSocket] # Funny they don't uniq
```

After requiring:
```
DRb::DRbProtocol.instance_variable_get(:@protocol)
[DRb::DRbUNIXSocket]
DRb::DRbProtocol.add_protocol(DRb::DRbTCPSocket)
[DRb::DRbUNIXSocket]
```

agrare:
@jrafanie yeah if we do decide to do this we should probably do this in core.  If setting the URI in the client's `start_service` would fix it though maybe we don't need this /cc @mkanoor

Fryguy:
@agrare I just noticed this is still open...should we get this in?

agrare:
@Fryguy I seem to remember while we were doing this the reporter wasn't able to reproduce it anymore to confirm that this fixed it.

Fryguy:
@agrare Oh interesting...what do you want to do then?

mkanoor:
@agrare @Fryguy 
We could not come to a conclusion on why DRb decided to switch to using TCP instead of pipes. The reporter was in a mixed test environment. Since we don't have any conclusive evidence we should not do this hack and patch DRb.

agrare:
@mkanoor :+1: closing

PR #239, opened 2018-10-02
expose auth_key to automate for AuthTokens

sseago:
v2v functionality needs auth_key available in automate.

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_auth_token.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_auth_token.rb
@@ -1,0 +1,5 @@
+module MiqAeMethodService
+  class MiqAeServiceAuthToken < MiqAeServiceAuthentication
+    expose :auth_key
+  end
+end
--- a/spec/service_models/miq_ae_service_auth_token_spec.rb
+++ b/spec/service_models/miq_ae_service_auth_token_spec.rb
@@ -1,0 +1,5 @@
+describe MiqAeMethodService::MiqAeServiceAuthToken do
+  it "#auth_key" do
+    expect(described_class.instance_methods).to include(:auth_key)
+  end
+end


aufi:
@fdupont-redhat

lib/miq_automation_engine/service_models/miq_ae_service_auth_token.rb
```diff
@@ -0,0 +1,5 @@
+module MiqAeMethodService
+  class MiqAeServiceAuthToken < MiqAeServiceAuthentication
+    expose :auth_key
```
gmcculloug: Columns for the db table are automatically exposed, but we are explicitly blocking this one: https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb#L6

May need to discuss if there are other options.  @fdupont-redhat Can you add a reference to where it is being used.

fdupont-redhat: This is meant to be used to connect to the conversion host in OpenStack. The private key is stored conversion host's provider auth_key. The code that connects via SSH to the conversion host needs it.

It was supposed to be done in Automate, but with https://github.com/ManageIQ/manageiq-schema/pull/281 merged and backported to Hammer, I tend to prefer extending https://github.com/ManageIQ/manageiq/pull/18033 with an additional PR that moves the SSH connection logic in the ConversionHost class, which has access to the provider's auth_key.

The other benefit of moving the SSH connection to ConversionHost is that I can leverage the MiqSshUtil gem and reduce the specific code in ConversionHost.

sseago: @gmcculloug Note that the blocked password field is explicitly exposed in https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_authentication_rhsm.rb -- so there is precedent for this approach.

aufi: OpenStack (and propably other providers) use Authentication also to store keys for connection to OpenStack. I believe that this kind of usage should be exposed to Automate for V2V and also use cases when Automate script wants connect to OSP using the key (different than connection via APIs).

Could we possibly make auth_key exposed only in a subclass (or role) SSH key for OpenStack (or RHV etc.)?

mansam: Hi @gmcculloug, any thoughts on the last couple of comments?

PR #240, opened 2018-10-02
[HAMMER] Update i18n catalog for hammer

mzazrivec:


locale/en/ManageIQ_AutomationEngine.po
```diff
@@ -0,0 +1,29 @@
...
+"Report-Msgid-Bugs-To: \n"
+"PO-Revision-Date: 2018-10-02 16:32+0200\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: English\n"
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=n != 1;\n"
+"\n"
```
gmcculloug: @mzazrivec Should the headers in these files look more like: https://github.com/ManageIQ/manageiq/blob/master/locale/en/manageiq.po

`FIRST AUTHOR` and `Last-Translator` look like they should be updated or removed.  And `Project-Id-Version` is set to `1.0.0` in some places and `master` in others.

How do we do this across other repos?  I would expect it to be consistent.

mzazrivec: These files are automatically generated, i.e. these headers are filled in automatically.

Once the actual translations from Zanata land, those headers will point to actual names.

mzazrivec:
I uploaded current hammer catalogs (i.e. what's in this PR) to Zanata for the initial translations to commence. I created this (and other similar PRs) so that we're consistent with what's right now in Zanata.

PR #250, opened 2018-10-11
[WIP] convert_value_based_on_datatype won't take args unless they're downcase

d-m-u:
We call automate with the array of objects that a custom button is run on, which looks something like this: ```Integer::ID``` and ... integer needs to be downcased so it doesn't hit https://github.com/ManageIQ/manageiq-automation_engine/blob/f294b5636db07acfea58391230a7e8e41be73b1b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb#L572 and barf. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1628224


d-m-u:
@miq-bot add_reviewer @mkanoor

d-m-u:
@miq-bot add_label bug, hammer/yes

d-m-u:
@miq-bot add_reviewer @tinaafitz

PR #251, opened 2018-10-11
Use Open3.popen3 instead of Open4

NickLaMuro:
**DISCLAIMER**:  There is no rational for me doing this beyond the fact I don't like gems sitting around we don't really need.  We don't need to do this, so feel free to close if you (the reviewer) would prefer to avoid the risk, because this is definitely a heavily used portion of the Automation code base.  More rational can be found below.

* * *

Basically, I noticed this `open4` gem when bundling `manageiq` by chance, and thought "What the F is this doing..."

Turns out it is only used in [two spots](https://github.com/search?l=Ruby&q=org%3AManageIQ+Open4&type=Code):

- Here (this repo)
- The Cockpit worker (main repo)

Both really aren't making any specific use of `Open4` specific functionality as far as I can tell, and since that gem hasn't been updated in the past 4 years, seems like it isn't the best thing to be relying on.

Tackled this one first since the `manageiq` change would require this one removed first, and if this relies on some feature of `Open4` that I am missing, it is a complete blocker for the other change.

Detailed Notes
--------------

As far as I can tell, We aren't making use of any of the differentiating features of the `Open4` project, and seeing as it hasn't had any updates for 4+ years now, I think it makes sense that we try and avoid using it if it isn't providing anything useful for us.

I am personally not an expert on `Open4`, but here is what I have determined is "different" about it:

- It uses `fork` + `exec` instead of `spawn` (in what I think is kind of a hacky approach)
- It will automatically convert the `STDIN`/`STDOUT` to `sync = true` for you
- It will Marshall your output for you from the forked process 

Currently, we don't make use of the process Marshalling here (we only look at the exit code from the process), and due to my unfamiliarity with, I am unsure if the `fork` + `exec` approach is somehow a key component to allowing Automate to function.  My guess would be no, but I don't have a good way of verifying, besides running the specs, which seem to work fine with this change.

The `STDIN`/`STDOUT` only sees relevant if we want streaming output, but seeing as they are updated within the code in `DrbRemoteInvoker`, so I can only assume that that would overwrite anything that is configured in `Open4`.

**Implementation**

This does require being slightly more explicit in the block with `Open3`, as the value of the block now becomes the return value assigned to `status`, instead of a shared return value for block/non-block forms in `Open4` (the non-block form return value very much matches `Open4`'s, with just a re-ordering of the arguments, for what that is worth).

Links
-----

* Places used in the `ManageIQ` organization:  https://github.com/search?l=Ruby&q=org%3AManageIQ+Open4&type=Code
* Ruby doc info for `Open3.popen3`:  https://ruby-doc.org/stdlib-2.4.1/libdoc/open3/rdoc/Open3.html#method-c-popen3

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -191,13 +191,13 @@ module MiqAeEngine
     private_class_method :invoke_inline_ruby
 
     def self.run_method(cmd)
-      require 'open4'
+      require 'open3'
       rc = nil
       threads = []
       method_pid = nil
       begin
-        status = Open4.popen4(*cmd) do |pid, stdin, stdout, stderr|
-          method_pid = pid
+        status = Open3.popen3(*cmd) do |stdin, stdout, stderr, wait_thread|
+          method_pid = wait_thread.pid
           yield stdin if block_given?
           stdin.close
           threads << Thread.new do
@@ -207,6 +207,7 @@ module MiqAeEngine
             stderr.each_line { |msg| $miq_ae_logger.error "Method STDERR: #{msg.strip}" }
           end
           threads.each(&:join)
+          wait_thread.value
         end
         rc  = status.exitstatus
         msg = "Method exited with rc=#{verbose_rc(rc)}"


jrafanie:
I don't think there was a requirement to use open4 vs. open3.  

```
commit addb10af13940e9f88808f33b2a9aeac811b99f9
Author: <REDACTED>
Date:   Mon Oct 11 23:30:00 2010 +0000

    Rewrite Automate Method Launch (MiqAeMethod.invoke_inline_ruby) using Open4 gem
    BugzID: 10316
```

I'm guessing we saw Open4 used somewhere and decided to try it out.  I don't  see why we can't use Open3.  I know automate uses RC/exit code and output but I don't know that we exercise all the code paths in tests to be confident.

cc @mkanoor

gmcculloug:
cc @billfitzgerald0120 Please test

gmcculloug:
Changes look right based on the block form example described here: https://www.rubydoc.info/stdlib/open3/Open3.popen3

gmcculloug:
@NickLaMuro I was finally able to test this on an appliance in several scenarios including successful and failed service provisioning attempts and I am seeing all the same results/logging with this changes as with `open4` it all looks good.

Just wanted to check with you on the `[POC]'` title.  I'm willing to merge this if you think it is ready.

NickLaMuro:
@gmcculloug Thanks!  That is good to here.  I didn't have any way of doing that, so the only validation that I did have that it was working was the tests passing (which I did have them fail on my originally when I did some things wrong, so that was good).

Unless you think that what you did for tests was not thorough enough, I am completely fine removing that label and merging this.

PR #253, opened 2018-10-12
Add conversion_host= method to MiqAeServiceServiceTemplateTransformationPlanTask

fdupont-redhat:
This PR adds a `conversion_host=` method to allow setting the conversion host of a task from Automate.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1634029

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
@@ -31,8 +31,8 @@ module MiqAeMethodService
     def conversion_host=(conversion_host)
       raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
       ar_method do
-        return wrap_results(@object.conversion_host = nil) if conversion_host.nil?
-        @object.conversion_host = ConversionHost.find_by(:id => conversion_host.id)
+        conversion_host = ConversionHost.find_by(:id => conversion_host.id) if conversion_host
+        @object.conversion_host = conversion_host
         @object.save
       end
     end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
@@ -29,7 +29,7 @@ module MiqAeMethodService
     end
 
     def conversion_host=(conversion_host)
-      raise ArgumentError, "conversion_host must be a MiqAeServiceConversionHost" unless conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
+      raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
       ar_method { wrap_results(@object.conversion_host = ConversionHost.find(conversion_host.id)) }
     end
   end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
@@ -27,5 +27,15 @@ module MiqAeMethodService
         wrap_results(@object.transformation_destination(source_obj.object_class.find(source_obj.id)))
       end
     end
+
+    def conversion_host=(conversion_host)
+      raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
+
+      ar_method do
+        @object.conversion_host = conversion_host && conversion_host.instance_variable_get("@object")
+        _log.info "Setting Conversion Host on #{@object.class.name} id:<#{@object.id}> to #{@object.conversion_host.inspect}"
+        @object.save
+      end 
+    end 
   end
 end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
@@ -31,6 +31,6 @@ module MiqAeMethodService
     def conversion_host=(conversion_host)
       raise ArgumentError, "conversion_host must be a MiqAeServiceConversionHost" unless conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
       ar_method { wrap_results(@object.conversion_host = ConversionHost.find(conversion_host.id)) }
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@miq-bot add-label transformation, enhancement, hammer/yes

lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
```diff
@@ -27,5 +27,15 @@ def transformation_destination(source_obj)
         wrap_results(@object.transformation_destination(source_obj.object_class.find(source_obj.id)))
       end
     end
+
+    def conversion_host=(conversion_host)
+      raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
+
+      ar_method do
+        @object.conversion_host = conversion_host && conversion_host.instance_variable_get("@object")
```
mkanoor: @fdupont-redhat Do we need this check since you are doing the same check in raise ArgumentError?

gmcculloug: We should avoid accessing the internal properties of the service model.  In the past I have used the fact that the service model exposes the same ID as the active_record model did a `find` on that object.

Like this: https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_vm.rb#L9

lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
```diff
@@ -27,5 +27,15 @@ def transformation_destination(source_obj)
         wrap_results(@object.transformation_destination(source_obj.object_class.find(source_obj.id)))
       end
     end
+
+    def conversion_host=(conversion_host)
```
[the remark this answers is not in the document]
fdupont-redhat: @gmcculloug I've followed existing code in miq_ae_service_service_template.rb where `owner=` and `group=` use a parameter named identically to the method name. But, I can change it if you prefer.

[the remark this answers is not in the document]
gmcculloug: Fair point, I can live with it either way.

fdupont-redhat:
@gmcculloug Yes, I accept a nil conversion host, as it resets the task as unassigned.

lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
```diff
@@ -27,5 +27,10 @@ def transformation_destination(source_obj)
...
+      raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
+      ar_method { wrap_results(@object.conversion_host = ConversionHost.find(conversion_host.id)) }
```
gmcculloug: Now that we are back to supporting `nil` as a potential host parameter the `conversion_host.id` call will fail with `undefined method 'id' for nil:NilClass`

Which highlights the need for tests on the PR.

lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_task.rb
```diff
@@ -27,5 +27,14 @@ def transformation_destination(source_obj)
         wrap_results(@object.transformation_destination(source_obj.object_class.find(source_obj.id)))
       end
     end
+
+    def conversion_host=(conversion_host)
+      raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
+      ar_method do
+        return wrap_results(@object.conversion_host = nil) if conversion_host.nil?
```
gmcculloug: You should not need `wrap_results` here and I'm also confused why you have to save `@object` below but not here.  Seems you would want to go through the same steps.

Would something like this work?
```ruby
def conversion_host=(conversion_host)
  raise ArgumentError, "conversion_host must be nil or a MiqAeServiceConversionHost" unless conversion_host.nil? || conversion_host.kind_of?(MiqAeMethodService::MiqAeServiceConversionHost)
  ar_method do
    conversion_host = ConversionHost.find_by(:id => conversion_host.id) if conversion_host
    @object.conversion_host = conversion_host
    @object.save
  end
end
```

fdupont-redhat: Neat. I like the override of `conversion_host`.

spec/service_models/miq_ae_service_service_template_transformation_plan_task_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  describe "#conversion_host=" do
+    before(:each) do
+      @conversion_host = FactoryGirl.create(:conversion_host)
+    end
+
+    it "removes the conversion host if arg is nil" do
+      task.conversion_host = @conversion_host
+      method = "$evm.root['#{@ae_result_key}'] = $evm.root['service_template_transformation_plan_task'].conversion_host = nil"
```
gmcculloug: @fdupont-redhat I would prefer we avoid the overhead of instantiation the engine to test this, you should be able to create the active_record and service_model objects and test directly.

There are several tests that can be used as an example.  Here's one: https://github.com/ManageIQ/manageiq-automation_engine/blob/master/spec/service_models/miq_ae_service_orchestration_stack_spec.rb

fdupont-redhat: @gmcculloug Done. My previous tests were inspired by https://github.com/ManageIQ/manageiq-automation_engine/blob/master/spec/service_models/miq_ae_service_vm_spec.rb#L75.

spec/service_models/miq_ae_service_service_template_transformation_plan_task_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe MiqAeMethodService::MiqAeServiceServiceTemplateTransformationPlanTask do
+  let(:conversion_host_1) { FactoryGirl.create(:conversion_host) }
+  let(:service_conversion_host_1) { MiqAeMethodService::MiqAeServiceConversionHost.find(conversion_host_1.id) }
```
gmcculloug: `service_conversion_host_1` is never used.

simaishi:
Hammer backport details:

```
$ git log -1
commit 736eba51486c296d95eaf7b5294e4335675f690a
Date:   Wed Oct 17 08:36:16 2018 -0400

    Merge pull request #253 from fdupont-redhat/add_conversion_host_method
    
    Add conversion_host= method to MiqAeServiceServiceTemplateTransformationPlanTask
    
    (cherry picked from commit e2329f72fea82c4440189baa223d2bb30bf04f8b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1634029
```

PR #263, opened 2018-10-26
Expose validate_conversion_hosts for TransformationPlanRequest

fdupont-redhat:
This PR is the automation-engine counterpart of https://github.com/ManageIQ/manageiq/pull/18135. It will allow Automate to check the request and deny it if no conversion host is available.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1640816


fdupont-redhat:
@miq-bot add-label transformation, enhancement, hammer/yes

JPrause:
@miq-bot add_label blocker

lib/miq_automation_engine/service_models/miq_ae_service_service_template_transformation_plan_request.rb
```diff
@@ -1,6 +1,7 @@
 module MiqAeMethodService
   class MiqAeServiceServiceTemplateTransformationPlanRequest < MiqAeServiceServiceTemplateProvisionRequest
     expose :source_vms, :association => true
+    expose :validate_conversion_hosts, :association => true
```
gmcculloug: This is fine but I'm realizing that none of these methods actually return an association so all of the `:associations => true` should be removed.

Even `source_vms` (which should likely be renamed at some point) is only returning IDs not actually objects which is what you would expect for an association, and the naming.

@fdupont-redhat Please remove the `:association = true` for `validate_conversion_hosts` line and I'll create a follow-up PR to resolve the other ones so we do not conflate this PR with both enhancements and code cleanup.

cc @bzwei

bzwei: Thanks @gmcculloug

fdupont-redhat: @gmcculloug Done.

simaishi:
Hammer backport details:

```
$ git log -1
commit 2b84cbe753cb9f8ba1febc4ec35455f7eac017a7
Date:   Tue Oct 30 10:39:11 2018 -0400

    Merge pull request #263 from fdupont-redhat/bz1640816
    
    Expose validate_conversion_hosts for TransformationPlanRequest
    
    (cherry picked from commit a0a7e527a936ebdaae411b06e00cba3795da6f47)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1640816
```

PR #265, opened 2018-10-30
Add automate method for external Ansible Job/Workflow template.

lfu:
Allow user to create an automate method type of AnsibleTemplate which allows for the selection of an external Tower Provider and a Job/Workflow Template.

Depends on https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/136.
Includes https://github.com/ManageIQ/manageiq/pull/18160.

Fix https://bugzilla.redhat.com/show_bug.cgi?id=1591797


mkanoor:
@lfu can you fix the rescue issue as reported by rubocop

mkanoor:
@lfu How is the code coverage on the new files. The earlier files we had 100% coverage we want to ensure the same with the new files

mkanoor:
@lfu 
Do you need a new constant here
https://github.com/ManageIQ/manageiq/blob/9822d52744420c753c05b664d0d3d7cefa33edf9/app/models/miq_ae_method.rb#L20

So that the UI can show the new method type, be it an ansible job template or a playbook

lfu:
@mkanoor Added ManageIQ/manageiq#18160.

lfu:
screen shot 2018-11-05 at 11 12 12

app/models/mixins/miq_ae_ansible_method_mixin.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def wait_for_method(task_id)
+    task = MiqTask.wait_for_taskid(task_id)
+    raise MiqAeException::Error, task.message unless task.status == "Ok"
+    process_result
+  ensure
+    reset
+  end
+
+  def check_task_status(task_id)
```
mkanoor: @lfu 
Can we break this method into 3 separate small functions

app/models/mixins/miq_ae_ansible_method_mixin.rb
```diff
@@ -0,0 +1,136 @@
+module MiqAeAnsibleMethodMixin
```
gmcculloug: I think this makes more sense as a base class than a mixin.  You should turn this into a class like `MiqAeEngine::MiqAeAnsibleMethodBase` and sub-class the Playbook and Template classes off of that.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -50,109 +23,13 @@ def execute
...
-    end
-
-    def current_info
-      list = %w(namespace class instance message method)
-      list.each.with_object({}) { |m, hash| hash[m] = @workspace.send("current_#{m}".to_sym) }
-    end
-
-    def build_options_hash
+    def build_playbook_options_hash
+      build_options_hash
```
mkanoor: @lfu can we call the method build_options_hash and call the parent classes build_options_hash and then add the extra stuff. 
```
def build_options_hash
   super.tap do |config_info|
      config_info[:hosts] = resolved_hosts
   end
end
```

lfu:
@miq-bot add_label hammer/no

PR #269, opened 2018-11-08
[WIP] Fix an issue with uninitialized constant ManageIQ::Automate::System::CommonMethods.

lfu:
Fix an issue with uninitialized constant ManageIQ::Automate::System::CommonMethods when an automate method calls /System/Request/InspectMe as an embedded method.

The error was thrown out because InspectMe ran methods on /System/CommonMethods/Utils/LogObject before the class was loaded.

@miq-bot add_label bug

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -242,6 +242,7 @@ module MiqAeEngine
     def self.bodies_and_line_numbers(obj, aem)
       embeds = []
       embedded_methods(obj.workspace, aem, embeds, aem.fqname)
+      embeds.reverse!
       embeds << {:data => aem.data, :fqname => aem.fqname}
       code_start = 0
       script_info = {}
--- a/spec/engine/miq_ae_method_spec.rb
+++ b/spec/engine/miq_ae_method_spec.rb
@@ -259,6 +259,7 @@ describe MiqAeEngine::MiqAeMethod do
               Level3.log_me
             end
           end
+          Level3.log_me
         RUBY
       end
       let(:level2_embeds) { ['/Shared/Methods/Level3'] }


gmcculloug:
@lfu @mkanoor Can you explain what about InspectMe is causing the issue.  Originally the loading order was discussed here (https://github.com/ManageIQ/manageiq-automation_engine/pull/207#issuecomment-415458472) and I'm not sure how much we want to change it.  It sounds like InspectMe is not written to be an embedded method but we are trying to use it that way which might be the issue.

mkanoor:
@gmcculloug 
Normally the embedded methods are pure classes with no extra lines of code outside the class definition. In @lfu 's case she has embedded the InspectMe method, which is not an ideal candidate for embedding. The InspectMe method has a method call outside the class definition that calls the main method. InspectMe method includes the logging embedded method, which has not been loaded at this point causing a undefined symbol error

The methods are loaded in the order they are found along the method tree.

This PR reverses the order of the embedded methods so that any direct invocation would work.

``` ruby
# This class embeds class B
class A
   def self.method_A
         B.method_B
   end
end
# Directly call the method_A
A.method_A


# This class embeds Class C
class B
   def self.method_B
       C.method_C
   end
end
# The line below is the problem because C has not been loaded yet
B.method_B

class C
   def self.method_C
       puts 1
   end
end
```

So if the load order is
Class B
Class C
Class A

it breaks because C is not defined when B executes the method_B class

But if you change the load order in reverse
Class C
Class B
Class A
It works

gmcculloug:
Discussed this issue with @mkanoor and @tinaafitz and we decided to leave this WIP for the moment as we would like to collect more information on the usage of the nested embedded methods.

The issue that concerns me is the `InspectMe` was never intended to be called as an embedded method so reversing the order that methods are loaded today potentially break existing configurations.

For example:
Method A
-- Method B  (Util class)
-- Method C  Call to Util class outside of a method

The above configuration works today, but if we reverse the ordering it would break.  Also, visually it feels better that the methods load in the order they are defined.

Make this wip for now so we can evaluate the approach a little more.

lfu:
Close it for now.

PR #271, opened 2018-11-14
Adding a domain lock check for automate import.

pkomanek:
Removes an option to import into the locked domain in automate with local or git import. This check runs only for enviroment with selected user. If there is no user it means that the import is from the console or bin/update script. This PR is part of the https://github.com/ManageIQ/manageiq-ui-classic/pull/4912.

### DIFF
--- a/app/models/miq_ae_yaml_import.rb
+++ b/app/models/miq_ae_yaml_import.rb
@@ -15,6 +15,9 @@ class MiqAeYamlImport
   def import
     if @options.key?('import_dir') && !File.directory?(@options['import_dir'])
       raise MiqAeException::DirectoryNotFound, "Directory [#{@options['import_dir']}] not found"
+    elsif !User.current_user.nil? && @options['zip_file'] && domain_locked?(@options['import_as'])
+      # raises exception only for a local import into the locked domain
+      raise MiqAeException::DomainNotAccessible, 'locked domain'
     end
     start_import(@options['preview'], @domain_name)
   end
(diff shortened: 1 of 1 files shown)


mkanoor:
@pkomanek 
Is this to prevent 2 concurrent imports from stepping on each other or
Is it that a locked domain (GIT, ManageIQ or any user locked) should not allow imports?

Also we allow for updates to a GIT Repo from the UI would this prevent that too?

app/models/miq_ae_yaml_import.rb
```diff
@@ -15,10 +15,17 @@ def initialize(domain, options)
...
       raise MiqAeException::DirectoryNotFound, "Directory [#{@options['import_dir']}] not found"
+    elsif !User.current_user.nil? && domain_locked?
+      raise MiqAeException::DomainNotAccessible, 'locked domain'
     end
     start_import(@options['preview'], @domain_name)
   end
 
+  def domain_locked?
+    dom = domain_files('*').first.split('/').first
+    (@options['import_as'] && MiqAeDomain.find_by(:name => @options['import_as'])&.contents_locked?) || (dom && MiqAeDomain.find_by(:name => dom)&.contents_locked?)
```
gmcculloug: @pkomanek As discussed it would be good to move this line into a separate method with a descriptive name, otherwise it is difficult to parse at first glance.

tinaafitz:
@mkanoor Please review.

PR #274, opened 2018-12-11
Refactoring automate system domain list to include legacy domains

d-m-u:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1651497. 

We're only currently listing MIQ in the list of domains to be reset, and it should include the RH domain. 

Jason and I and GM discussed pulling out the domain list itself into a separate method. 

The UI call is here: https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/miq_ae_tools/_import_export.html.haml#L73.

### DIFF
--- a/app/models/miq_ae_datastore.rb
+++ b/app/models/miq_ae_datastore.rb
@@ -130,7 +130,7 @@ module MiqAeDatastore
     seed_default_namespace
   end
 
-  def self.reset_domain(datastore_dir, domain_name, tenant)
+  private_class_method def self.reset_domain(datastore_dir, domain_name, tenant)
     _log.info("Resetting domain #{domain_name} from #{datastore_dir}")
     ns = MiqAeDomain.find_by_fqname(domain_name)
     ns.destroy if ns
@@ -140,6 +140,10 @@ module MiqAeDatastore
       ns.update_attributes!(:source   => MiqAeDomain::SYSTEM_SOURCE, :enabled => true,
                             :priority => MANAGEIQ_PRIORITY) if ns
     end
+  end
+
+  private_class_method def self.reset_domains(domains)
+    domains.each { |domain| reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant) }
   end
 
   def self.seed_default_namespace
@@ -161,38 +165,50 @@ module MiqAeDatastore
   def self.reset_to_defaults
     saved_attrs = preserved_attrs_for_domains
 
-    reset_domains_from_legacy_directory
-    reset_domains_from_vmdb_plugins
+    reset_all_domains
 
     restore_attrs_for_domains(saved_attrs)
     reset_default_namespace
     MiqAeDomain.reset_priorities
   end
 
-  def self.reset_domains_from_legacy_directory
+  private_class_method def self.domains
+    legacy_domains + plugin_domains
+  end
+
+  private_class_method def self.default_domains
+    domains.select(&:system?)
+  end
+
+  private_class_method def self.legacy_domains
     domain_files = DATASTORE_DIRECTORY.join('*', MiqAeDomain::DOMAIN_YAML_FILENAME)
-    Dir.glob(domain_files).each do |domain_file|
-      domain_name = File.basename(File.dirname(domain_file))
-      reset_domain(DATASTORE_DIRECTORY, domain_name, Tenant.root_tenant)
+    domains = Dir.glob(domain_files).map do |domain_file|
+      Vmdb::Plugins::AutomateDomain.new(File.dirname(domain_file))
     end
+    filter_domains(domains)
   end
 
-  def self.reset_domains_from_vmdb_plugins
-    datastores_for_reset.each do |domain|
-      reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant)
-    end
+  private_class_method def self.plugin_domains
+    filter_domains(Vmdb::Plugins.automate_domains)
   end
 
-  def self.datastores_for_reset
-    domains = Vmdb::Plugins.automate_domains
+  private_class_method def self.filter_domains(domains)
     return domains unless Rails.env.test? && ENV["AUTOMATE_DOMAINS"]
 
     filter = ENV["AUTOMATE_DOMAINS"].split(",")
     domains.select { |i| filter.include?(i.name) }
   end
 
+  private_class_method def self.domain_names
+    domains.collect(&:name)
+  end
+
   def self.default_domain_names
-    Vmdb::Plugins.system_automate_domains.collect(&:name)
+    default_domains.collect(&:name)
+  end
+
+  private_class_method def self.reset_all_domains
+    reset_domains(domains)
   end
 
   def self.seed
@@ -207,7 +223,7 @@ module MiqAeDatastore
         _log.info "Seeding... Complete"
       end
     end
-    _log.info "Reseting domain priorities at startup..."
+    _log.info("Resetting domain priorities at startup...")
     MiqAeDomain.reset_priorities
   end
 
--- a/spec/models/miq_ae_datastore_spec.rb
+++ b/spec/models/miq_ae_datastore_spec.rb
@@ -94,8 +94,24 @@ describe MiqAeDatastore do
     expect(MiqAeMethod.count).to         eq(3)
   end
 
-  it ".default_domain_names" do
-    expect(MiqAeDatastore.default_domain_names).to include("ManageIQ")
+  describe ".default_domain_names" do
+    describe "vmdb plugin domains" do
+      it "includes MIQ" do
+        expect(MiqAeDatastore.default_domain_names).to include('ManageIQ')
+      end
+    end
+
+    describe "legacy domains" do
+      it "includes system domain" do
+        stub_const("MiqAeDatastore::DATASTORE_DIRECTORY", ManageIQ::AutomationEngine::Engine.root.join('spec/models/miq_ae_datastore/data/automate_domain_list_tests'))
+        expect(MiqAeDatastore.default_domain_names).to include('drew_test_system')
+      end
+
+      it "doesn't include non-system domain" do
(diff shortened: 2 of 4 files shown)


d-m-u:
@miq-bot add_label bug

app/models/miq_ae_datastore.rb
```diff
@@ -170,17 +170,16 @@ def self.reset_to_defaults
...
 
   def self.reset_domains_from_vmdb_plugins
-    datastores_for_reset.each do |domain|
-      reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant)
-    end
+    datastores_for_reset.each { |domain| reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant) }
+  end
+
+  def self.domain_names
+    files = DATASTORE_DIRECTORY.join('*', MiqAeDomain::DOMAIN_YAML_FILENAME)
```
gmcculloug: I would suggest that we make two private methods `plugin_domain_names` and `legacy_domain_names` and this method would becomes `plugin_domain_names + legacy_domain_names`

app/models/miq_ae_datastore.rb
```diff
@@ -256,4 +254,17 @@ def self.path_includes_domain?(path, options = {})
     nsd, = ::MiqAeEngine::MiqAePath.split(path, options)
     MiqAeNamespace.find_by_fqname(nsd, false) != nil
   end
+
+  def self.plugin_domain_names
+    datastores_for_reset
```
Fryguy: Minor, but this method doesn't return domain names, it return domain objects.  So later when you do `plugin_domain_names.each { |domain| ...domain.datastores_path...` it reads funny because I would expect Strings to come out of a method with that name.

app/models/miq_ae_datastore.rb
```diff
@@ -170,17 +170,15 @@ def self.reset_to_defaults
...
   def self.reset_domains_from_vmdb_plugins
-    datastores_for_reset.each do |domain|
-      reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant)
-    end
+    plugin_domain_names.each { |domain| reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant) }
+  end
+
+  def self.domain_names
+    plugin_domain_names + legacy_domain_names
   end
```
Fryguy: I think this needs to be taken one step further and the` default_domain_names` method needs to change (line 192).  That method is the one that defines the "system" domains for the UI, but it is ignoring the legacy domains.  The refactoring here should mostly help with getting those legacy domains into that list.

app/models/miq_ae_datastore.rb
d-m-u: Sure. May I just omit this method entirely since it's pretty superfluous?

Fryguy: Yeah probably.

app/models/miq_ae_datastore.rb
```diff
@@ -256,4 +254,13 @@ def self.path_includes_domain?(path, options = {})
     nsd, = ::MiqAeEngine::MiqAePath.split(path, options)
     MiqAeNamespace.find_by_fqname(nsd, false) != nil
   end
+
+  def self.legacy_domains
+    domain_files = DATASTORE_DIRECTORY.join('*', MiqAeDomain::DOMAIN_YAML_FILENAME)
+    Dir.glob(domain_files).map do |domain_file|
+       Vmdb::Plugins::AutomateDomain.new(File.dirname(domain_file))
```
Fryguy: Oh that's a smart way to do it :+1: ... also minor 3 space indent on this line

d-m-u:
😭 Madhu says I can't use the spec/data in the engine repo so it has to be this fun, which is going to fail...

Fryguy:
> Madhu says I can't use the spec/data in the engine repo 

Why not?  @mkanoor ?

d-m-u:
@miq-bot remove_label wip

d-m-u:
@Fryguy @mkanoor @bdunne 
anyone want to review this, please? it's not blocker but i'd like to get it off my plate

spec/models/miq_ae_datastore_spec.rb
```diff
@@ -94,8 +94,20 @@ def setup_version_xml(v)
     expect(MiqAeMethod.count).to         eq(3)
   end
 
-  it ".default_domain_names" do
-    expect(MiqAeDatastore.default_domain_names).to include("ManageIQ")
+  describe "#default_domain_names" do
```
bdunne: This should be `".default_domain_names"` since it's a class method

spec/models/miq_ae_datastore_spec.rb
```diff
@@ -94,8 +94,20 @@ def setup_version_xml(v)
...
+    context "vmdb plugin domains" do
```
bdunne: I think the `context` & `it ".default_domain_names"` is redundant, should the context be removed and the `it` changed to `vmdb plugin domains`?

spec/models/miq_ae_datastore_spec.rb
```diff
@@ -94,8 +94,20 @@ def setup_version_xml(v)
...
+      it ".default_domain_names" do
+        expect(MiqAeDatastore.default_domain_names).to include('ManageIQ')
+      end
+    end
+
+    context "legacy domains" do
```
bdunne: same here

app/models/miq_ae_datastore.rb
```diff
@@ -170,17 +170,15 @@ def self.reset_to_defaults
...
-      reset_domain(DATASTORE_DIRECTORY, domain_name, Tenant.root_tenant)
-    end
+    legacy_domains.each { |domain_name| reset_domain(DATASTORE_DIRECTORY, domain_name, Tenant.root_tenant) }
   end
 
   def self.reset_domains_from_vmdb_plugins
-    datastores_for_reset.each do |domain|
-      reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant)
-    end
+    plugin_domain_names.each { |domain| reset_domain(domain.datastores_path.to_s, domain.name, Tenant.root_tenant) }
```
Fryguy: - `plugin_domain_names` should really be renamed to `plugin_domains` since it returns objects.
- `legacy_domains` also returns object, and so could use `domain.datastores_path.to_s` instead of `DATASTORE_DIRECTORY`

So given both of those, these 2 methods become identical, and we can refactor them away.

app/models/miq_ae_datastore.rb
```diff
@@ -192,7 +190,7 @@ def self.datastores_for_reset
   end
 
   def self.default_domain_names
-    Vmdb::Plugins.system_automate_domains.collect(&:name)
+    domain_names.collect(&:name)
```
Fryguy: I think this is supposed to be system domain (including those that are legacy), not all domains.

d-m-u:
@miq-bot remove_label wip

d-m-u:
@bdunne @Fryguy can you all please look again if you get a chance? thanks.

Fryguy:
Moving the code can be done in a separate PR.  Have you verified that all of the deleted methods are not longer used?

The only other thing that I think you should do in this PR is mark any new methods that aren't being used externally as private methods.

d-m-u:
@Fryguy I checked on the deleted methods and I think we're good, can you please look?

d-m-u:
@Fryguy @bdunne can I please get eyes on this?

d-m-u:
🐢 🐌 🐿

app/models/miq_ae_datastore.rb
```diff
@@ -24,6 +24,10 @@ def self.temp_domain
     "#{TEMP_DOMAIN_PREFIX}-#{SecureRandom.uuid}"
   end
 
+  def self.datastore_directory
```
Fryguy: Not a fan of wrapping a method around a constant.  Particularly because we probably don't want this to be a public method, but this seems like the only issue.

app/models/miq_ae_datastore.rb
```diff
@@ -161,38 +165,50 @@ def self.seed_default_namespace
...
     filter = ENV["AUTOMATE_DOMAINS"].split(",")
     domains.select { |i| filter.include?(i.name) }
   end
 
+  private_class_method def self.domain_names
+    domains.collect(&:name)
+  end
+
   def self.default_domain_names
-    Vmdb::Plugins.system_automate_domains.collect(&:name)
```
Fryguy: @d-m-u Aside, I think this is the last caller of Vmdb::Plugins.system_automate_domains, and if so, can you please open a PR to delete it from Vmdb::Plugins in core?

simaishi:
@d-m-u Can this be `hammer/yes`?
@dmetzger57 FYI, this is labeled as `refactoring`

d-m-u:
ya. 
@miq-bot add_label hammer/yes

d-m-u:
it was refactoring to fix a bug, 's both bug n refactor.
refactorfixing?

simaishi:
yeah, I understand. Just wanted to give @dmetzger57 a heads up as we discussed about giving a bit more attention to any PR that refactor code.

dmetzger57:
Thanks @simaishi  for the heads up on refactoring aspect in a back port, will advise QE to give this one a little more attention.

simaishi:
Hammer backport details:

```
$ git log -1
commit e0cceef7553b9c97d85543d85b6d03b2e71bd249
Author: Jason Frey <fryguy9@gmail.com>
Date:   Wed Jan 16 17:57:50 2019 -0500

    Merge pull request #274 from d-m-u/refactoring_domain_reset
    
    Refactoring automate system domain list to include legacy domains
    
    (cherry picked from commit 06ed15d7bb29dcb823eee3356727fe3d086a5501)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686054
```

app/models/miq_ae_datastore.rb
d-m-u: https://github.com/ManageIQ/manageiq/pull/18526

PR #287, opened 2019-01-15
Don't overwrite message in create_automation_object

d-m-u:
Creation of a custom button to request call_instance_with_message is overwriting the message in  create_automation_object and we shouldn't do that. We can have one message stored in ```attrs``` and the other in ```options[:message]``` and they are distinct entities and should remain so. This appears to have been wrong for _quite a while_ (https://github.com/ManageIQ/manageiq-automation_engine/commit/9c5233b21455f5cc06af132e67f30c998f149c9b) (and also slightly less relevant but still worth noting, https://github.com/ManageIQ/manageiq/commit/227afe61061296de7d90f743fdf54a9bb0e13232). This worked in most cases because we don't usually have a message attribute in the attr value pairs, just the request message (which defaults to ```create```), so the message in the attrs was nil.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1651099

This is coming back into the edit screen for the button, note the fact that the messages are different, one hasn't been overwritten by the other. 
screen shot 2019-01-15 at 9 59 36 am

PR #288, opened 2019-01-15
Use lazy evaluation

d-m-u:
Just a silly refactor to use lazy eval instead of creating all the things every time, whether we need them or not. 

Per https://github.com/ManageIQ/manageiq-automation_engine/pull/287#discussion_r248051489. 

It's three commits cause the second is fixing all the rubocop issues with this file and the third is removing two ```allow_any_instance_ofs``` that we're not actually using.

### DIFF
--- a/spec/miq_ae_engine_spec.rb
+++ b/spec/miq_ae_engine_spec.rb
@@ -1,14 +1,15 @@
 describe MiqAeEngine do
   include Spec::Support::AutomationHelper
 
+  let(:user) { FactoryBot.create(:user_with_group) }
+  let(:domain) { 'SPEC_DOMAIN' }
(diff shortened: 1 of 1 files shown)


d-m-u:
@miq-bot add_label cleanup
@miq-bot assign @gmcculloug 
@miq-bot add_reviewer @mkanoor

d-m-u:
The allow_any_instance of was here loooooong before I was.

d-m-u:
@miq-bot add_reviewer @bdunne

d-m-u:
@miq-bot add_label developer

d-m-u:
@bdunne will you 👀 please? 
I lied, only 2/4 AAIOs can be taken out

d-m-u:
🐢 🐌 🌷

PR #292, opened 2019-02-04
 Redirect writable_storages to writable_accessible_storages

d-m-u:
"Redirect the existing writable_storages method exposed to automate to call writable_accessible_storages.  This would avoid having to update automate methods to use the new method.  From automate I'm not sure a user would need to distinguish between writable_storages and writable_accessible_storages." - The Estimable Mr Greg McCullough himself

Per https://bugzilla.redhat.com/show_bug.cgi?id=1668020

specs: https://github.com/ManageIQ/manageiq-content/pull/506

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_host.rb
@@ -9,7 +9,7 @@ module MiqAeMethodService
 
     expose :storages,              :association => true
     expose :read_only_storages
-    expose :writable_storages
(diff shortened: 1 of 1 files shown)


gmcculloug:
Marking as `hammer/no` since this was resolved on older versions by marking the storages as read-only to avoid schema migrations.  This change is only required on master.

d-m-u:
Specs here: https://github.com/ManageIQ/manageiq-content/pull/506

d-m-u:
@miq-bot remove_label wip

PR #296, opened 2019-02-12
Drop 2.3.x from test matrix, add 2.5.x

djberg96:
Drop Ruby 2.3.6 from the test matrix and add 2.5.3. All the cool kids are doing it:

https://github.com/ManageIQ/manageiq/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 rvm:
 - 2.4.5
-- 2.3.8
+- 2.5.3
 sudo: false
 cache: bundler
 env:


gmcculloug:
@jrafanie I don't see many other repos that have made this switch yet, but I think it makes sense.  Just looking for another opinion.  Asking since you were involved in https://github.com/ManageIQ/manageiq/pull/18313.

Also, this pesky issue of trying to get tests to finish.

jrafanie:
yeah, if I were to guess, I'd imagine it's something with the drb threads and the remote invoker that's hanging:

https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb

I'll try to debug that when I clear away some of the rails upgrade stuff i'm looking at.

djberg96:
@jrafanie Is this an issue with 2.5.x, or does this just need to be restarted?

jrafanie:
> @jrafanie Is this an issue with 2.5.x, or does this just need to be restarted

@djberg96 I'd assume it's a 2.5 issue if this has happened more than once.   The referenced code is reaching into the internals of drb and has been painful to maintain as we upgrade rubies.

jrafanie:
Additionally, https://github.com/ManageIQ/manageiq-automation_engine/pull/298 should fix the other failure I saw after https://github.com/ManageIQ/manageiq/pull/18476 is merged

jrafanie:
@djberg96 the PR was merged on manageiq, want to close/open this PR?  It should then fail on 2.5 with the error #298 fixes.

djberg96:
@jrafanie kicked it, i'm guessing the current failure is the expected one.

jrafanie:
yay, it's 💚

djberg96:
@gmcculloug Good to go?

PR #298, opened 2019-02-21
Ruby 2.5 won't escape tilde as per RFC3986 Section 2.3

jrafanie:
This landed in ruby 2.5.0:
https://github.com/ruby/ruby/commit/e1b432754553423008a14d39d0901eabc99e7ddb

Hopefully we weren't relying on this because ruby 2.4 will escape ~ but
2.5 will not.

Looks like this was added in this commit back on old pre-manageiq:

```
commit b87b70c826c1e20badf2b9afd25acfa385b54583
Author: Gilly
Date:   Tue Jun 25 17:24:26 2013 -0400

    Fully escape attributes and their values passed to automate.

    hash2query method now escapes both key and value as well as escaping any
    non-alphanumeric.

    Bug 971596, 976901
    Issue #553
```

cc @gmcculloug @bdunne

### DIFF
--- a/spec/miq_ae_uri_spec.rb
+++ b/spec/miq_ae_uri_spec.rb
@@ -16,9 +16,9 @@ describe MiqAeEngine::MiqAeUri do
   end
 
   it "escape non-ASCII Numeric characters" do
-    hash = {"test://dev?lab$~" => "OU=serverbuildingtest, OU=dev2"}
+    hash = {"test://dev?lab$" => "OU=serverbuildingtest, OU=dev2"}
     query = described_class.hash2query(hash)
-    expect(query).to eq("test%3A%2F%2Fdev%3Flab%24%7E=OU%3Dserverbuildingtest%2C%20OU%3Ddev2")
+    expect(query).to eq("test%3A%2F%2Fdev%3Flab%24=OU%3Dserverbuildingtest%2C%20OU%3Ddev2")
 
     result_hash = described_class.query2hash(query)
     expect(result_hash).to eq(hash)


jrafanie:
Note, this will fix a failure on ruby 2.5 that passes on 2.4

spec/miq_ae_uri_spec.rb
```diff
@@ -16,9 +16,9 @@
   end
 
   it "escape non-ASCII Numeric characters" do
-    hash = {"test://dev?lab$~" => "OU=serverbuildingtest, OU=dev2"}
+    hash = {"test://dev?lab$" => "OU=serverbuildingtest, OU=dev2"}
     query = described_class.hash2query(hash)
```
bdunne: Can we delete this and use `Hash#to_query` instead?  Is it just for handling the AR Objects?

gmcculloug: It handles AR objects: https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_uri.rb#L10

Plus returns `nil` if the hash is empty, where to_query would be a blank string.  Not sure if there is logic relying on that result.

But I could see refactoring that method to take advantage of `to_query`.

djberg96: Ok, let's refactor another time. I vote to merge, and we can refactor another time.

spec/miq_ae_uri_spec.rb
```diff
@@ -16,9 +16,9 @@
...
-    expect(query).to eq("test%3A%2F%2Fdev%3Flab%24%7E=OU%3Dserverbuildingtest%2C%20OU%3Ddev2")
+    expect(query).to eq("test%3A%2F%2Fdev%3Flab%24=OU%3Dserverbuildingtest%2C%20OU%3Ddev2")
```
bdunne: Should the `%7E` be replaced with a `~` instead of removing it from the test string?

gmcculloug: Then the test would need to be conditional based on the version of ruby since it is processed differently in 2.4 and 2.5 as seen here: https://travis-ci.org/ManageIQ/manageiq-automation_engine/builds/496782188

gmcculloug: @bdunne What do you think?  I'm ok with dropping it as it seems like we would just be testing ruby at that point.

bdunne: :+1: I'm okay with dropping it.  I didn't know if it was an important use case

PR #293, opened 2019-02-11
Don't set User.current_user

d-m-u:
We have three assignments to User.current_user inside this repo and they're problematic because it isn't getting properly reset between threads. We should be using the ```with_user``` block for this. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1671563

PR #302, opened 2019-03-12
Set current user for exposed methods in automate.

lfu:
So when the methods run from an automate script, it would have the current user set properly.
 
Followup of https://github.com/ManageIQ/manageiq-automation_engine/pull/293.

Related to https://github.com/ManageIQ/manageiq/pull/18545.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -164,13 +164,17 @@ module MiqAeMethodService
         self.association = method_name if options[:association]
         define_method(method_name) do |*params|
           method = options[:method] || method_name
-          ret = object_send(method, *params)
+          ret = User.with_user(self.class.workspace&.ae_user) { object_send(method, *params) }
           return options[:override_return] if options.key?(:override_return)
           options[:association] ? wrap_results(self.class.filter_objects(ret)) : wrap_results(ret)
         end
       end
     end
     private_class_method :expose
+
+    def self.workspace
+      MiqAeEngine::MiqAeWorkspaceRuntime.current || MiqAeEngine::DrbRemoteInvoker.workspace
+    end
 
     def self.wrap_results(results)
       ar_method do
--- a/spec/service_models/miq_ae_service_service_template_provision_request_spec.rb
+++ b/spec/service_models/miq_ae_service_service_template_provision_request_spec.rb
@@ -15,15 +15,26 @@ describe MiqAeMethodService::MiqAeServiceServiceTemplateProvisionRequest do
     MiqAeEngine.instantiate("/EVM/AUTOMATE/test1?ServiceTemplateProvisionRequest::service_template_provision_request=#{@service_template_provision_request.id}", @user)
   end
 
-  it "#approve" do
-    approver = 'wilma'
-    reason   = "Why Not?"
-    method   = "$evm.root['#{@ae_result_key}'] = $evm.root['service_template_provision_request'].approve('#{approver}', '#{reason}')"
-    @ae_method.update_attributes(:data => method)
-    expect(MiqRequest).to receive(:find).with(@service_template_provision_request.id.to_s)
-      .and_return(@service_template_provision_request)
-    expect(@service_template_provision_request).to receive(:approve).with(approver, reason).once
-    expect(invoke_ae.root(@ae_result_key)).to be_truthy
+  context "#approve" do
+    it "gets called by method" do
+      approver = 'wilma'
+      reason   = "Why Not?"
+      method   = "$evm.root['#{@ae_result_key}'] = $evm.root['service_template_provision_request'].approve('#{approver}', '#{reason}')"
+      @ae_method.update_attributes(:data => method)
+      expect(MiqRequest).to receive(:find).with(@service_template_provision_request.id.to_s).and_return(@service_template_provision_request)
+      expect(@service_template_provision_request).to receive(:approve).with(approver, reason).once
+      expect(invoke_ae.root(@ae_result_key)).to be_truthy
+    end
+
+    it "sets current user" do
+      user = FactoryBot.create(:user_with_group)
+      ws = double(:ae_user => user)
+      allow(service_service_template_provision_request.class).to receive(:workspace).and_return(ws)
+      approver = FactoryBot.create(:user_miq_request_approver)
+
+      expect(User).to receive(:with_user).with(user)
+      service_service_template_provision_request.approve(approver, "for test")
+    end
   end
 
   it "#user_message" do


lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -159,12 +159,13 @@ def self.expose(*args)
...
       raise ArgumentError, "cannot have :method option if there is more than one method name specified" if options.key?(:method) && args.length != 1
 
+      user = User.current_user
       args.each do |method_name|
         next if method_name.to_sym == :id
         self.association = method_name if options[:association]
         define_method(method_name) do |*params|
           method = options[:method] || method_name
-          ret = object_send(method, *params)
+          ret = User.with_user(user) { object_send(method, *params) }
```
gmcculloug: The `user` variable is set outside of the `define_method` block and therefore will be capture when `expose` is run, which is at service model creation time.  The `user` captured at that time would always be used regardless of the user context when the method is run on an instance of the class.

Here is a simple example showing the results:
```ruby
class ABC
  def self.expose(method_name)
    user = $current_user
    define_method(method_name) do
      puts "User: #{user}"
    end
  end
end

$current_user = "Initial Name"
ABC.expose :test_method

ABC.new.test_method
# => User: Initial Name

$current_user = "Updated Name"
ABC.new.test_method
# => User: Initial Name
```

I believe what you are really trying to do is to run in the user context when the instance method is run, right?  For that you would need to determine the `user` inside of the `define_method` block.

mkanoor: @gmcculloug 
Its the same user, the user is not changing under the covers. Instead the problem is that Drb runs requests coming from Automate Methods in different threads and those threads don't have the Thread Local Storage set for current user. This closure approach of wrapping the user before the DRb method runs will work. Once Automate starts it runs with the given user context we don't allow them to change it.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -159,12 +159,13 @@ def self.expose(*args)
       options = args.last.kind_of?(Hash) ? args.pop : {}
       raise ArgumentError, "cannot have :method option if there is more than one method name specified" if options.key?(:method) && args.length != 1
 
+      user = User.current_user
```
mkanoor: @lfu We need to raise an error here if the user is not set.

mkanoor: It should not happen but in case someone stumbles upon it from a simulation or other code paths we should be able to catch it.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
gmcculloug: My point is the class is going to get created at some point for a process and once it is created it will remain until the process ends.  A request from a different user that runs in the same process would be run in the context of the user when the class was created.  It is very possible the class would be created in a system context and callers to the method defined in `expose` would always have a `nil` user context.

Let me know if you want to discuss.

mkanoor: @gmcculloug I was under the impression that the classes are going to get loaded for every Automate Request coming in since the service models are going to call the expose method.
If that is not true then we should just fetch the workspace from the DRb service front like its done here

https://github.com/ManageIQ/manageiq-automation_engine/blob/7b39eb1d8a4dba982bd33781a77bab8d728e5bf9/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_rbac.rb#L39

The expose method should fetch the workspace and the workspace has a ae_user object that can be used with a with_user

gmcculloug: The `expose` method is called once during the class creation [1] and the method defined by `define_method` would be created at that time.  When the user calls the dynamically defined method, for example `ext_management_system` from the link below, the `user` value would already be set from creation time.

[1] https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_vm_or_template.rb#L14

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -164,14 +164,18 @@ def self.expose(*args)
         self.association = method_name if options[:association]
         define_method(method_name) do |*params|
           method = options[:method] || method_name
-          ret = object_send(method, *params)
+          ret = User.with_user(self.class.workspace.ae_user) { object_send(method, *params) }
```
gmcculloug: From inside a class method you want just `workspace.ae_user`.  The `self.class` part would reference the class of the `MiqAeMethodService::MiqAeServiceModelBase` class which is `Class` and would not respond to the `workspace` method.

lfu: Seems the exposed method runs as an instance method. Without ```self.class``` it would run into an error:
```
[----] W, [2019-03-13T14:04:55.710817 #12084:f06f54]  WARN -- : Error during substitution: undefined local variable or method `workspace' for #<MiqAeMethodService::MiqAeServiceMiqEvent:0x0000000008449b40>
```

lfu:
I have done the end-to-end test following these steps:
1. create two users A and B
2. create a generic service C
3. login as user A and order service C
4. login as user B and order service C
5. check automation.log for lines like:
```
[----] I, [2019-03-14T10:25:06.288314 #24494:7b4f58]  INFO -- : MiqAeEvent.build_evm_event >> event=<"request_created"> inputs=<{"EventStream::event_stream"=>1000000003483, :event_stream_id=>1000000003483}>
[----] I, [2019-03-14T10:26:14.234806 #24674:7b4f58]  INFO -- : User [B] with current group ID [1000000000005] name [EvmGroup-administrator]
```
When the process started with user B, that message should say ```User [B]``` instead of ```User [A]```.

gmcculloug:
@lfu When I mentioned tests I was thinking of spec tests, not manual testing.  Sorry if that was confusing.

lfu:
I have not figured out how to test it with specs. Any ideas?

gmcculloug:
Looking over the code change I would say the test would validate that `User.with_user` is called with the expected user when a method defined by `define_method` is called.  You could stub `workspace` to return an object to access the user.

spec/service_models/miq_ae_service_service_template_provision_request_spec.rb
```diff
@@ -15,15 +15,27 @@ def invoke_ae
...
+      expect(invoke_ae.root(@ae_result_key)).to be_truthy
+    end
+
+    it "sets current user" do
+      user = FactoryBot.create(:user_with_group)
+      ws = double(:ae_user => user)
+      allow(service_service_template_provision_request.class).to receive(:workspace).and_return(ws)
+      approver = FactoryBot.create(:user_miq_request_approver)
+
+      User.current_user = user
```
gmcculloug: This line does not appear to be needed.  Otherwise it looks good.

tinaafitz:
@miq-bot add_label bug

simaishi:
@lfu https://github.com/ManageIQ/manageiq-automation_engine/pull/293 is `hammer/yes`. Is this needed in hammer branch too?

lfu:
@miq-bot add_label hammer/yes, bug

simaishi:
Hammer backport details:

```
$ git log -1
commit f81ad32d6743e10616094821b16e2770ce93a4dd
Date:   Mon Mar 18 17:34:31 2019 -0400

    Merge pull request #302 from lfu/current_user
    
    Set current user for exposed methods in automate.
    
    (cherry picked from commit 5ccdf7db5d5e73f27b7385e02594818073a70efe)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1728706
```

PR #303, opened 2019-03-12
Support Export/Import of Domain with Ansible Method 

pkomanek:
Adding the support for Export/Import of Domain with Ansible Method. Right now if you try to export domain with the playbook type method and put it on a second appliance it could be invalid. It is because the exported method contains the IDs of related objects like a repository or playbook and the second appliance don't have to contain all related objects or may contain the same objects with a different IDs. This could be a problem and the imported playbook method on the second appliance could be invalid. This PR contains one conversion of these playbook attributes IDs to their related object names for the export and a second conversion inside the import, which finds the related objects by their names and then converts these names back to their IDs. This PR is related to the second one bellow with a new flash message for the import datastore with an invalid playbook method.

links:
-----------
* New flash message PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/5331
* New factory for the embedded ansible cloud credential: https://github.com/ManageIQ/manageiq/pull/18644
* Based on https://bugzilla.redhat.com/show_bug.cgi?id=1677575


pkomanek:

app/models/miq_ae_yaml_export.rb
```diff
@@ -187,6 +188,13 @@ def write_method_attributes(method_obj, export_file_hash)
     write_export_file(export_file_hash)
   end
 
+  def convert_playbook_attributes(envelope_hash)
+    opt = envelope_hash['object']['attributes']['options']
+    opt[:repository_id] = AnsibleRepositoryController.model.find(opt[:repository_id]).name
+    opt[:playbook_id]   = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Playbook.find(opt[:playbook_id]).name
+    opt[:credential_id] = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Credential.find(opt[:credential_id]).name
```
lfu: What happens if some of these objects have been removed?

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,37 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
...
+      raise MiqAeException::AttributeNotFound, ruby_method_file_name
+    end
+  end
+
+  def remove_playbook_attributes_names(options)
+    options.delete(:repository_name)
+    options.delete(:playbook_name)
+    options.delete(:credential_name)
+    options.delete(:vault_credential_name)
+    options.delete(:cloud_credential_name)
```
lfu:
```
options.except!(:repository_name, :playbook_name, :credential_name, :vault_credential_name, :cloud_credential_name)
```

app/models/miq_ae_yaml_export.rb
```diff
@@ -187,6 +188,29 @@ def write_method_attributes(method_obj, export_file_hash)
...
+    options[:vault_credential_name] = ae_manager::VaultCredential.find_by(:id => options[:vault_credential_id])&.name if options[:vault_credential_id]
+    options[:cloud_credential_name] = ae_manager::CloudCredential.find_by(:id => options[:cloud_credential_id])&.name if options[:cloud_credential_id]
+  end
+
+  def remove_playbook_attributes_ids(options)
+    options.delete(:repository_id)
+    options.delete(:playbook_id)
+    options.delete(:credential_id)
+    options.delete(:vault_credential_id)
+    options.delete(:cloud_credential_id)
```
lfu:
```
options.except!(:repository_id, :playbook_id, :credential_id, :vault_credential_id, :cloud_credential_id)
```

lfu:
@pkomanek Please take a look at the Travis failure.

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,29 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
     if method_attributes['location'] == 'inline'
       data = load_method_ruby(ruby_method_file_name)
       method_yaml.store_path('object', 'attributes', 'data', data) if data
+    elsif method_attributes['location'] == 'playbook'
+      convert_playbook_attributes(method_attributes['options'], ruby_method_file_name)
+      method_attributes['options'].except!(:repository_name, :playbook_name, :credential_name, :vault_credential_name, :cloud_credential_name)
```
lfu: Better to clear these name attributes inside method ```convert_playbook_attributes```.

pkomanek: @lfu thanks for the review. I moved the name removing to the `convert_playbook_attributes` and added 1 more test for the situation, where is the playbook method created, then the playbook is deleted and the export is called.

app/models/miq_ae_yaml_export.rb
```diff
@@ -177,6 +177,7 @@ def write_all_methods(ns_fqname, class_obj)
 
   def write_method_attributes(method_obj, export_file_hash)
     envelope_hash = setup_envelope(method_obj, METHOD_OBJ_TYPE)
+    convert_playbook_attributes(envelope_hash) if method_obj.location == "playbook"
```
mkanoor: @lfu @pkomanek what happens if this is the new workflow method

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,29 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
...
     track_stats('method', method_obj)
     method_obj ||= add_method(class_obj, method_yaml) unless @preview
     method_obj
   end
 
+  def convert_playbook_attributes(options, ruby_method_file_name)
+    ae_manager = ManageIQ::Providers::EmbeddedAnsible::AutomationManager
+    begin
+      options[:repository_id]       = AnsibleRepositoryController.model.find_by(:name => options[:repository_name]).id.to_s if options[:repository_name]
+      options[:playbook_id]         = ae_manager::Playbook.find_by(:name => options[:playbook_name]).id.to_s                if options[:playbook_name]
```
mkanoor: @pkomanek During import how do we tell the user that the repository could not be found of playbook_name could not be found, the error message has to to tell you what exactly is missing

app/models/miq_ae_yaml_export.rb
lfu: This PR works only for ```playbook``` method.

The [PR](https://github.com/ManageIQ/manageiq/pull/18160) to add job/workflow template method is still pending. We can address that in a separate PR later.

tinaafitz:
hi @pkomanek Is this PR still WIP?

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,74 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
...
+                AnsibleRepositoryController.model
+              when 'playbook'
+                ae_manager::Playbook
+              when 'credential'
+                ae_manager::Credential
+              when 'vault_credential'
+                ae_manager::VaultCredential
+              when 'cloud_credential'
+                ae_manager::CloudCredential
+              end
```
lfu: The definition of ```klass``` and ```ae_manager``` can be moved into ```convert_playbook_attribute```.

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,74 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
...
+      raise MiqAeException::AttributeNotFound, error_msg
+    end
+    options.except!(:repository_name, :playbook_name, :credential_name, :vault_credential_name, :cloud_credential_name)
+  end
+
+  def convert_playbook_attribute(options, klass, attr, invalid_attributes)
+    related_obj = klass.find_by(:name => options["#{attr}_name".to_sym])
+
+    if related_obj
+      options["#{attr}_id".to_sym] = related_obj.id.to_s
```
lfu: Any reason for ```to_s```?

app/models/miq_ae_yaml_import.rb
```diff
@@ -222,13 +222,74 @@ def process_method(class_obj, ruby_method_file_name, method_yaml)
...
+    else
+      invalid_attributes << attr.to_s
```
lfu: Same here.

app/models/miq_ae_yaml_import.rb
pkomanek: @lfu If you create a new Playbook Method via UI, these IDs are stored as a string values. Also when I tried to change them to numbers instead of strings, at least 1 page stopped working. Editing page did not load these values, so I think that UI assume strings there.

spec/models/miq_ae_yaml_import_export_spec.rb
```diff
@@ -674,6 +674,86 @@ def assert_class_without_methods(export_options, import_options)
...
+      export_model(@manageiq_domain.name, @export_options)
+      reset_and_import(@export_dir, @manageiq_domain.name, @import_options)
+      check_counts(@domain_counts)
+      check_playbook_attributes(true)
+    end
+
+    def check_playbook_attributes(skip_playbook)
+      opt = MiqAeMethod.find_by(:name => method.name).attributes['options']
+      if skip_playbook
+        expect(opt[:playbook_id]).to(be_nil)
```
lfu: Please delete the ```()``` after ```to```.
```
expect(opt[:playbook_id]).to be_nil
...
expect(opt[:repository_id]).to eq(repo.id.to_s)

```

spec/models/miq_ae_yaml_import_export_spec.rb
```diff
@@ -738,19 +738,19 @@ def assert_class_without_methods(export_options, import_options)
...
-      expect(opt[:verbosity]).to(eq(''))
-      expect(opt[:cloud_credential_id]).to(eq(cloud_cred.id.to_s))
-      expect(opt[:execution_ttl]).to(eq('256'))
-      expect(opt[:hosts]).to(eq('localhost'))
-      expect(opt[:log_output]).to(eq("on_error"))
-      expect(opt[:become_enabled]).to(be_falsey)
+      expect(opt[:repository_id]).to eq(repo.id.to_s)
+      expect(opt[:credential_id]).to eq(cred.id.to_s)
+      expect(opt[:vault_credential_id]).to eq(vault_cred.id.to_s)
+      expect(opt[:verbosity]).to eq('')
```
lfu: ```expect(opt[:verbosity]).to be_empty```

PR #304, opened 2019-03-13
MiqAeMethodCopy: fix for broken copy method

pkomanek:
Adding the missing option hash for a new `MiqAeMethod` created by `MiqAeMethodCopy#copy`. 

Steps for QA:
----------------------
(Option hash is used for playbook attributes)
1) Go to Automation -> Automate -> Explorer
2) Create a new `MiqAeMethod` with the `playbook` type
3) Set some playbook attributes (for example: repository).
4) Save this new method
5) Copy this method

* **old result:** New method with blank playbook attributes
* **new result:** New method with all playbook attributes

### DIFF
--- a/app/models/miq_ae_method_copy.rb
+++ b/app/models/miq_ae_method_copy.rb
@@ -55,6 +55,7 @@ class MiqAeMethodCopy
     validate
     create_method
     @dest_method.inputs << add_inputs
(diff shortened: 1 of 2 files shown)


pkomanek:
@miq-bot add_label bug

PR #305, opened 2019-03-18
Remove references to MiqPassword

Fryguy:
@gmcculloug Please review.  This gets us closer to removing MiqPassword from manageiq-gems-pending.


Fryguy:
cc @jrafanie

PR #306, opened 2019-03-19
Physical server provisioning via StateMachine

miha-plesko:
With this commit we make sure that AE engine calculates correct    entrypoint when PhysicalServer provisioning request is created,     namely:
    
```
PhysicalInfrastructure/PhysicalServer/Lifecycle/Provisioning
```
    
Furthermore, we enhance PhysicalServer AE service model with `ems_type`     method that is used for execution routing inside StateMachine     (i.e. via messages).

Depends on https://github.com/ManageIQ/manageiq/pull/18573
Related to https://github.com/ManageIQ/manageiq-content/pull/516

@miq-bot assign @gmcculloug 

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
@@ -10,10 +10,12 @@ module MiqAeEngine
       vm_retire_task
       service_retire_task
       orchestration_stack_retire_task
+      physical_server_provision_task
       platform_category
     ).freeze
     CLOUD          = 'cloud'.freeze
     INFRASTRUCTURE = 'infrastructure'.freeze
+    PHYSICAL_INFRA = 'physicalinfrastructure'.freeze
     SERVICE        = 'service'.freeze
     UNKNOWN        = 'unknown'.freeze
 
@@ -70,7 +72,8 @@ module MiqAeEngine
         when 'orchestration_stack_retire' then %w(Orchestration Lifecycle Retirement)
         when 'configured_system_provision'
           obj.workspace.root['ae_provider_category'] = 'infrastructure'
-          %w(Configured_System Lifecycle Provisioning)
+          %w[Configured_System Lifecycle Provisioning]
+        when 'physical_server_provision' then %w[PhysicalServer Lifecycle Provisioning]
         end
       $miq_ae_logger.info("Request:<#{obj['request']}> Target Component:<#{obj['target_component']}> ")
       $miq_ae_logger.info("Target Class:<#{obj['target_class']}> Target Instance:<#{obj['target_instance']}>")
@@ -181,12 +184,23 @@ module MiqAeEngine
       case obj_name
       when "orchestration_stack", "orchestration_stack_retire_task"
         CLOUD
-      when "miq_request", "miq_provision", "vm_migrate_task", "vm_retire_task"
+      when "miq_request"
+        case prov_obj
+        when nil
+          nil
+        when PhysicalServerProvisionRequest
+          PHYSICAL_INFRA
+        else
+          vm_detect_category(prov_obj.source)
+        end
+      when "miq_provision", "vm_migrate_task", "vm_retire_task"
         vm_detect_category(prov_obj.source) if prov_obj
       when "service_retire_task"
         ""
       when "vm"
         vm_detect_category(prov_obj) if prov_obj
+      when "physical_server_provision_task"
+        PHYSICAL_INFRA
       else
         UNKNOWN
       end
--- a/spec/engine/miq_ae_builtin_method_spec.rb
+++ b/spec/engine/miq_ae_builtin_method_spec.rb
@@ -16,4 +16,55 @@ describe MiqAeEngine::MiqAeBuiltinMethod do
       expect { subject }.to raise_error(MiqAeException::StopInstantiation)
     end
   end
+
+  describe 'AE entrypoint calculation' do
+    let(:obj)       { double('OBJ', :workspace => workspace) }
+    let(:workspace) { double('WORKSPACE', :root => root) }
+    let(:root)      { {} }
+
+    context '/PhysicalInfrastructure/PhysicalServer/Lifecycle/Provisioning' do
+      before do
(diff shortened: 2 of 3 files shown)


miha-plesko:
Retrying add_label and assign:

@miq-bot assign @gmcculloug

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
```diff
@@ -182,11 +185,15 @@ def self.detect_category(obj_name, prov_obj)
       when "orchestration_stack", "orchestration_stack_retire_task"
         CLOUD
       when "miq_request", "miq_provision", "vm_migrate_task", "vm_retire_task"
+        return PHYSICAL_INFRA if (prov_obj&.options || {})[:request_type] == 'provision_physical_server'
```
gmcculloug: Instead of checking the options hash, could we verify they sub-class of object?  Same for the `miq_provision_task` case below.  Event though these are different objects, one request and one task, maybe we could pull this logic into a `physical_server?` method here to DRY up the logic.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
```diff
@@ -10,10 +10,12 @@ class MiqAeBuiltinMethod
       vm_retire_task
       service_retire_task
       orchestration_stack_retire_task
+      miq_provision_task
```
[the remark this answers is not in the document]
miha-plesko: As discussed on gitter I've now improved this part: `PhysicalServerProvisionTask` now resolves to `physical_server_provision_task` so that it's more obvious what we're provisioning. This also simplifies the IF statements below because task type is now unique.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
miha-plesko: Good point - turns out it's now even more simple because we've separated task name. For MiqRequest I've done it with case statement, please let me know if you find it readable enough. If not, I'll use auxilary function as you suggest.

lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
```diff
@@ -5,5 +5,9 @@ class MiqAeServicePhysicalServer < MiqAeServiceModelBase
     expose :turn_off_loc_led
     expose :power_on
     expose :power_off
+
+    def ems_type
+      ar_method { @object.ext_management_system.class.ems_type }
```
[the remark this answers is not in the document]
miha-plesko: Good point, done. I like your suggestion because we're now able to easier access that value even ouside of automation.

miha-plesko:
FYI Travis is failing because we need to merge core PR first https://github.com/ManageIQ/manageiq/pull/18573. @agrare gave us the thumbsup so @gmcculloug it's all yours :)

lib/miq_automation_engine/service_models/miq_ae_service_physical_server_provision_task.rb
```diff
@@ -0,0 +1,4 @@
+module MiqAeMethodService
+  class MiqAeServicePhysicalServerProvisionTask < MiqAeServiceMiqProvisionTask
+  end
+end
```
lfu: Is this file necessary? @gmcculloug

gmcculloug: No, automate will dynamically create service models and expose columns, virtual_columns and associations from the active_record model.  You only need to create these models now if you modify or expose additional functionality.

miha-plesko: Which I do - I extend `MiqAeServiceMiqProvisionTask` which have a bunch of custom methods like state_machine_status and set_option() that we need.

gmcculloug: The point being this model is empty so those methods are either being exposed automatically or pulled in from sub-classes.  If there is nothing defined here the empty class is not needed.

miha-plesko: I think it didn't work until I added the definition here because [this method](https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_miq_provision_task.rb#L3) was not defined  but let me try again tomorrow, probably I messed something up. Anyhow, thanks for pointing up, I had no idea service models are able to automatically inherit between themselves based on actual model relations.

gmcculloug: If you are interested in learning more check out https://github.com/ManageIQ/manageiq/pull/14465 and https://github.com/ManageIQ/manageiq-automation_engine/pull/308

miha-plesko: Surprisingly, you were right again 👍 Works like a charm without the file. Thanks for pointers to PRs

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
```diff
@@ -181,12 +184,23 @@ def self.detect_category(obj_name, prov_obj)
       case obj_name
       when "orchestration_stack", "orchestration_stack_retire_task"
         CLOUD
-      when "miq_request", "miq_provision", "vm_migrate_task", "vm_retire_task"
+      when "miq_request"
+        case prov_obj
+        when nil
+          UNKNOWN
```
lfu: With ```obj_name = 'miq_request'```, ```prov_obj = nil```,  seems old logic would return ```nil```. Is it right?

miha-plesko: Yes, but I think it's a bug and should be unknown. But you're right, I'll flip back to nil just to be safe.

gmcculloug:
Close/re-open to resolve failing tests.

miha-plesko:
I think it'll fail until core PR is merged where we define some factories

PR #308, opened 2019-03-20
Exposing service associations

d-m-u:
This is a refactor of the service model associations. We can expose all reflections (virtual and otherwise) [the same way we do with ```expose_class_attributes```](https://github.com/ManageIQ/manageiq-automation_engine/blob/5ccdf7db5d5e73f27b7385e02594818073a70efe/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb#L45).  

Automate has been able to auto-generate service models for active_record models for some time now (https://github.com/ManageIQ/manageiq/pull/14465).  This makes it possible to dynamically expose associations on models since the engine can handle converting any active_record models returned from the associations into the proper service models.

**Note**: The challenge here is that not all associations were properly exposed in the service models so there may be items that need to be addressed after this is merged.  I would expect two scenarios: 
1) Associations that were removed from active record but not updated on the service model.  (Likely not an issue as they would just no longer be defined.)
2) Active record methods that were added to service models with `:association => true` when they are not real associations or virtual associations.  These will not exist unless the service models are updated to expose it explicitly or the active_record models are update to define them as virtual associations.

This horror started with https://bugzilla.redhat.com/show_bug.cgi?id=1688900 (which is for hammer, and this cannot be backported, so the PR associated with the hammer version is here: https://github.com/ManageIQ/manageiq-automation_engine/pull/309)

If this even actually works n passes tests, I'd expect this to look different on hammer unless, and I quote, "... Dennis thinks it's the best thing since sliced bread..." (hint -- he won't.)

remaining action items for follow-up work:
miq_ae_service_vm_or_template.rb
miq_ae_service_network_router.rb
miq_ae_service_miq_request.rb
miq_ae_service_miq_provision_request.rb
miq_ae_service_manageiq-providers-cloud_manager.rb
miq_ae_service_manageiq-providers-cloud_manager-vm.rb
all still have exposed things that are not picked up by rails associations/virtual associations list


d-m-u:
@tinaafitz @mkanoor it should work!

d-m-u:
@miq-bot assign @gmcculloug

d-m-u:
I think GM's right and cloud_network and subnet should be virtual associations

d-m-u:
@jrafanie sorry but gm says he won't merge this until there are at least four deletion-related emojis on this pr along with your blessing.

d-m-u:
@miq-bot add_reviewer not_bdunne

d-m-u:
@miq-bot add_reviewer @jrafanie

spec/service_models/miq_ae_service_container_replicator_spec.rb
```diff
@@ -26,8 +26,4 @@
   it "#metrics" do
     expect(described_class.instance_methods).to include(:metrics)
   end
-
-  it "#metric_zones" do
-    expect(described_class.instance_methods).to include(:metric_zones)
-  end
```
jrafanie: how to even?

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -114,7 +122,7 @@ def self.create_service_model(ar_model)
     def self.dynamic_service_model_creation(ar_model, super_class)
       Class.new(super_class) do |klass|
         ::MiqAeMethodService.const_set(model_to_service_model_name(ar_model), klass)
-        expose_class_attributes(klass)
+        inherited(klass)
```
jrafanie: This is scary.  Sorry, I didn't notice this before.

Can we move this expose logic from inherited to a new method `expose_the_stuff` and call it here and from inherited?

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -115,6 +123,7 @@ def self.dynamic_service_model_creation(ar_model, super_class)
       Class.new(super_class) do |klass|
         ::MiqAeMethodService.const_set(model_to_service_model_name(ar_model), klass)
         expose_class_attributes(klass)
+        expose_class_associations(klass)
```
jrafanie: 👍

gmcculloug: Based on https://github.com/ManageIQ/manageiq-automation_engine/pull/308#discussion_r267967934 we wanted to create a new method that both `inherited` and `dynamic_service_model_creation` used.  

If we need to add more class initialization logic in the future we are just forcing it to be added in two place now.

jrafanie: yeah, if we are going to add more very soon, it might make sense to extract that logic to a method either now or when we do it.  I don't know if it's very soon but we've thought more was coming in the future and it never came so I'm not against holding off on extracting it to a method when we need a 3rd consumer.

gmcculloug:
It's getting really close, but I would like @mkanoor and @tinaafitz to review.  Also, I'm working with @d-m-u to make sure we do not break any misconfigured associations with this change.  The thought is that we can identify them now and address in a follow up.

d-m-u:
@gmcculloug I've taken out from this all the ones you'd id'd as "the method does exist, just not as an association, and therefore we should keep the reference in the service model file for this first round."

gmcculloug:
cc @pemcg

d-m-u:
@miq-bot remove_label wip

d-m-u:
@tinaafitz @mkanoor could I get you to please comment on this?

pemcg:
What we really should do is identify any that we think might be edge cases, and run object_walker against the new service model. Some of the methods, like https://github.com/ManageIQ/manageiq-automation_engine/pull/308/files#diff-f760dd15a46f8cd892797fc423fc5bdc still have a couple of associations in, but it's not clear to me why. That might be a likely edge case to test for example

d-m-u:
thansk @pemcg. so as GM wrote in the description of this PR, it does not take out all the previous associations because this code will only generate associations present in the list of real associations or virtual associations. we do not include active record methods with the association flag in this.

Fryguy:
I'd be for merging this now, and just reverting if it's a real problem.  Logically, the interface has not changed with this PR.  I am concerned about exposing too much, and thus making it harder to "undo" service models, but in reality, we've exposed a lot already.

PR #318, opened 2019-05-06
Add scope to list automate domains from ancestors tenants

lpichler:
this scope will be used in IU in detail of tenants to list related(ancestor) automate domains.
so report data api is in IU of detail of tenant:

`/report_data?name_scope=with_tenant&db_model=MiqAeDomain&tenant_id=3`
 (this is pretty vague - I have it here just example)

# Links
https://bugzilla.redhat.com/show_bug.cgi?id=1678124
https://github.com/ManageIQ/manageiq/issues/18734 - tracking issue

@miq-bot assign @gtanzillo 
@miq-bot add_label changelog/yes, enhancement

### DIFF
--- a/app/models/miq_ae_domain.rb
+++ b/app/models/miq_ae_domain.rb
@@ -27,6 +27,11 @@ class MiqAeDomain < MiqAeNamespace
 
   include TenancyMixin
 
+  def self.with_tenant(tenant_id)
+    tenant = Tenant.find(tenant_id)
+    where(:tenant_id => tenant.ancestor_ids + [tenant_id])
+  end
+
   def self.enabled
     where(:enabled => true)
   end
--- a/spec/miq_ae_domain_spec.rb
+++ b/spec/miq_ae_domain_spec.rb
@@ -1,101 +1,129 @@
 describe MiqAeDomain do
   include Spec::Support::AutomationHelper
 
-  let(:root_tenant) { Tenant.seed }
-  before do
-    @user = FactoryBot.create(:user_with_group)
-    EvmSpecHelper.local_miq_server
-    root_tenant
-    setup_model
-  end
+  let!(:root_tenant) { Tenant.seed }
 
-  def setup_model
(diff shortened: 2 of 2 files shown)


tinaafitz:
@mkanoor @lfu Please review.

app/models/miq_ae_domain.rb
```diff
@@ -27,6 +27,11 @@ class MiqAeDomain < MiqAeNamespace
 
   include TenancyMixin
 
+  def self.with_tenant(tenant_id)
+    tenant = Tenant.find(tenant_id)
```
lfu: What if ```tenant_id``` does not exist?

martinpovolny: well in that case it makes no sense to call this method w/o a valid tenant

the code would raise an exception now

@lfu: do you think that returning an empty `ActiveRecord::Relation` is a better solution?

martinpovolny:
ping @lpichler : this work is needed for a RFE an the UI work depends on it. Please, move it forward.

app/models/miq_ae_domain.rb
lpichler: @lfu @martinpovolny I think to throw exception is what we need here - because `tenant_id == true` is (as you mentioned ) something what is unexpected.

lfu: Sounds good to me.

martinpovolny:
@lfu : any further issues with the PR?

Thank you!

lpichler:
ping @lfu @tinaafitz

app/models/miq_ae_domain.rb
gmcculloug: @lfu Looks good, but should this be `with_tenants` since it potentially returns multiple tenants?

lfu: Ah, I think it is ```with_tenant``` with a single tenant id.

gmcculloug: Got it, thanks @lfu.

PR #322, opened 2019-05-13
Log warning if RH domain version on disk and db version differ

d-m-u:
The RedHat domain on disk version and the version in the db can differ. At the moment we only run this check via the UI after an upgrade if the worker role is on. We've had requests to log the information instead and it maybe makes more sense to do it as part of the plugin setup.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693362

### DIFF
--- a/config/initializers/version_check.rb
+++ b/config/initializers/version_check.rb
@@ -1,0 +1,4 @@
+rh_domain = MiqAeDomain.find_by(:name => "RedHat")
+if rh_domain.version != rh_domain.available_version
+  $log.warn "Current version - #{rh_domain.version}, Available version - #{rh_domain.available_version}"
+end


d-m-u:
@miq-bot add_reviewer @jrafanie 
@miq-bot add_reviewer @Fryguy 
@miq-bot assign @gtanzillo

d-m-u:
Oh silliness.
@miq-bot assign @tinaafitz

config/initializers/version_check.rb
```diff
@@ -0,0 +1,4 @@
+rh_domain = MiqAeDomain.find_by(:name => "RedHat")
```
jrafanie: Sorry, but hitting the db in an initializer will cause rake tasks that shouldn't need a db to need one... such as asset compilation.  😢

jrafanie: I would suggest we check the version before seeding the datastore:
https://github.com/ManageIQ/manageiq-automation_engine/blob/a13df793acd0096db25182a2e6750f5b6aedde42/app/models/miq_ae_datastore.rb#L214

d-m-u:
Closing because it's been determined we should have this in the last_startup logging which only requires changes in core repo.

PR #328, opened 2019-06-11
Support for opening URL from the UI through automate.

martinpovolny:
https://bugzilla.redhat.com/show_bug.cgi?id=1643331
(~~https://bugzilla.redhat.com/show_bug.cgi?id=1550002~~)

requires https://github.com/ManageIQ/manageiq-schema/pull/380  (merged)
https://github.com/ManageIQ/manageiq/pull/18849 (merged)

docs PR: https://github.com/ManageIQ/guides/pull/355

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_vm.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_vm.rb
@@ -1,5 +1,8 @@
 module MiqAeMethodService
   class MiqAeServiceVm < MiqAeServiceVmOrTemplate
+    require_relative "mixins/miq_ae_external_url_mixin"
+    include MiqAeExternalUrlMixin
+
     def remote_console_url=(url)
       object_send(:remote_console_url=, url, MiqAeEngine::DrbRemoteInvoker.workspace.ae_user.id)
     end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_cloud_tenant.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_cloud_tenant.rb
@@ -2,5 +2,7 @@ module MiqAeMethodService
   class MiqAeServiceCloudTenant < MiqAeServiceModelBase
     expose :update_cloud_tenant
     expose :delete_cloud_tenant
+    require_relative "mixins/miq_ae_external_url_mixin"
+    include MiqAeExternalUrlMixin
   end
 end
--- a/lib/miq_automation_engine/service_models/miq_ae_service_ext_management_system.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_ext_management_system.rb
@@ -4,6 +4,8 @@ module MiqAeMethodService
     include MiqAeServiceInflectorMixin
     require_relative "mixins/miq_ae_service_custom_attribute_mixin"
     include MiqAeServiceCustomAttributeMixin
+    require_relative "mixins/miq_ae_external_url_mixin"
+    include MiqAeExternalUrlMixin
 
     expose :to_s
     expose :authentication_userid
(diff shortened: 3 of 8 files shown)


gmcculloug:
@martinpovolny Please include the BZ in the PR description.  Thanks.

lib/miq_automation_engine/service_models/mixins/miq_ae_saved_url_mixin.rb
```diff
@@ -0,0 +1,7 @@
+module MiqAeSavedUrlMixin
+  extend ActiveSupport::Concern
+  def saved_url=(url)
```
gmcculloug: @martinpovolny If we are specifically using this for the custom button URL it would be better if that name reflected that.  The method name exposed here does not have to match the backend model method, but that usually makes things easier so we should seriously consider naming at each level.

gmcculloug:
@martinpovolny Is this ready to merge?

martinpovolny:
@gmcculloug : I need to add the mixin for the provider classes too. Looking into that now.

martinpovolny:
I wonder if I need to add the mixin to `MiqAeServiceManageIQ_Providers_CloudManager` or into `MiqAeServiceManageIQ_Providers_BaseManager` and in the latter then where does the class live?

gmcculloug:
Based on the `BUTTON_CLASSES` defined here:
https://github.com/ManageIQ/manageiq/blob/master/app/models/custom_button.rb#L41

It would get added here to support Providers:
https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_ext_management_system.rb

lib/miq_automation_engine/service_models/miq_ae_service_vm.rb
```diff
@@ -1,5 +1,8 @@
 module MiqAeMethodService
   class MiqAeServiceVm < MiqAeServiceVmOrTemplate
+    require_relative "mixins/miq_ae_external_url_mixin"
+    include MiqAeExternalUrlMixin
```
gmcculloug: @martinpovolny VM was not in the list provided in https://bugzilla.redhat.com/show_bug.cgi?id=1550002#c6.

If a user can set both `remote_console_url=` and `external_url=` what happens?

martinpovolny: @gmcculloug : in the UI I want to look for the `external_url` first and then as a fallback for the `remote_console_url` (SystemConsole entity). So that backwards compatibility is kept.

Like this:
https://github.com/ManageIQ/manageiq-ui-classic/pull/5688/files#diff-65512d4378cfa992fb57fbceaaf14c59R262

I do not think that we should do things differently for different entities. So moving forward I'd like to deprecate the `remote_console_url`.

Does that make sense to you?

martinpovolny: Alternatively I'd change the `remote_console_url` method to assign to the same place as `external_url` and add a migration to remove the column. But again, it can be done as a next step.

gmcculloug: My understand is the `remote_console_url` has a lot more support today for the specific task of setting up a proxy to support VM consoles, but it can also support what `external_url ` is trying to do.  As long as a combined field can support the two use-cases then I'm good.

martinpovolny: When I added this it was used for this single purpose.

The entity, that it write's to, the `SystemConsole`, that is used for the proxies.

But not this particular column or the particular method `remote_console_url=`

gmcculloug: Got it, let's create an issue to track the follow items, otherwise I'm good with this change.  👍

PR #331, opened 2019-06-13
Emses in tests should only be leaf classes otherwise they lack types

d-m-u:
And we should only be able to add ones with a type, so this fixes specs broken by https://github.com/ManageIQ/manageiq/pull/18842.


d-m-u:
@miq-bot add_label test

d-m-u:
@jrafanie want to review?

PR #333, opened 2019-06-18
Save playbook set_stats data for successive playbook methods.

lfu:
set_stats data in a playbook would be saved and passed to successive playbook methods via automate workspace.

Includes https://github.com/ManageIQ/manageiq/pull/18889

https://bugzilla.redhat.com/show_bug.cgi?id=1678135

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_ansible_method_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_ansible_method_base.rb
@@ -2,6 +2,8 @@ module MiqAeEngine
   class MiqAeAnsibleMethodBase
     include AnsibleExtraVarsMixin
 
+    ANSIBLE_STATS_PREFIX = 'ansible_stats_'.freeze
+    ANSIBLE_STATS_PREFIX_LEN = ANSIBLE_STATS_PREFIX.size
     METHOD_KEY_SUFFIX = "_ansible_method_task_id".freeze
     MIN_RETRY_INTERVAL = 1.minute
 
@@ -38,10 +40,20 @@ module MiqAeEngine
       @workspace.root['ae_state_started'].present?
     end
 
-    def process_result
+    def process_result(task = nil)
       @aw.reload
       @workspace.update_workspace(@aw.output) if @aw.output
+      ansible_stats_from_task(task).each { |k, v| @workspace.persist_state_hash[k] = v }
       reset
+    end
+
+    def ansible_stats_from_task(task)
+      stats = task.task_results&.dig('ansible_stats')
+      return {} unless stats
+
+      stats.each_with_object({}) do |(attr, val), obj|
+        obj["#{ANSIBLE_STATS_PREFIX}#{attr}"] = val
+      end
     end
 
     def reset
@@ -52,7 +64,7 @@ module MiqAeEngine
     def wait_for_method(task_id)
       task = MiqTask.wait_for_taskid(task_id)
       raise MiqAeException::Error, task.message unless task.status == "Ok"
-      process_result
+      process_result(task)
     ensure
       reset
     end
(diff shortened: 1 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_ansible_method_base.rb
```diff
@@ -38,9 +38,11 @@ def running_in_state_machine?
       @workspace.root['ae_state_started'].present?
     end
 
-    def process_result
+    def process_result(task = nil)
       @aw.reload
       @workspace.update_workspace(@aw.output) if @aw.output
+      ansible_stats = task.task_results&.dig('ansible_stats')
+      @workspace.update_workspace('state_vars' => {'ansible_stats' => ansible_stats}) if ansible_stats
```
mkanoor: @lfu the state vars typically store scalar data and not hashes.
This line might save us https://github.com/ManageIQ/manageiq-automation_engine/blob/b4f7c7a9931d9a70ba8c918be2c4e2c7436e259b/lib/miq_automation_engine/engine/miq_ae_engine.rb#L66 but you might want to test it out so that it can preserve the nested hashes

mkanoor:
@lfu how do we send out input parameters that are password types?

lfu:
@mkanoor Input parameters of password type work the same way as for other automate methods. They are encrypted in the log messages but decrypted before sending to Ansible Tower.

spec/miq_ae_playbook_method_spec.rb
```diff
@@ -202,7 +205,10 @@
       end
 
       it "state finishes succesfully" do
+        miq_task.task_results = {'ansible_stats' => {'stats_1' => 'testing', 'stats_2' => true}}
```
mkanoor: @lfu can you change stats_1 and stats_2 to var1 and var2 otherwise its confusing to see repeating strings

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_deserialize_workspace.rb
```diff
@@ -1,8 +1,8 @@
 module MiqAeEngine
   module MiqAeDeserializeWorkspace
     def update_workspace(hash)
```
mkanoor: @lfu not sure if you should be calling this function at all. It's for the Ansible Module which modifies the playbook.
In your case you should be just setting the state_vars directly and not calling this function at all.

gmcculloug: Agreed.  It was unclear why we were introducing safe navigation until I read @mkanoor response.

tinaafitz:
@gmcculloug Please review.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_ansible_method_base.rb
```diff
@@ -120,12 +132,22 @@ def current_info
...
       end
     end
 
+    def ansible_stats_from_ws
+      @workspace.persist_state_hash.each_with_object({}) do |(attr, val), obj|
+        if attr.start_with?(ANSIBLE_STATS_PREFIX)
+          obj[attr[ANSIBLE_STATS_PREFIX_LEN..-1]] = val
+        else
+          obj[attr] = val
+        end
```
gmcculloug: With this loop a state_var called `var1` and one called `ansible_stats_var1` will only result in one hash key being set, which is expected.  My concern is that we have no way to guarantee the order and therefore could have unpredictable results.

I would suggest we create two hashes and at the end of the method return the result of merging the `non_ansible_hash` into the `ansible_hash` which would ensure the non-Ansible keys win and gives a consistent result.

Something like: `ansible_hash.merge(non_ansible_hash)`

PR #338, opened 2019-07-15
Add service_vars to pass variables between catalog items in a bundle

lfu:
https://bugzilla.redhat.com/show_bug.cgi?id=1678136

@miq-bot add_label enhancement, hammer/no

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
@@ -92,6 +92,33 @@ module MiqAeMethodService
       MiqAeEngine::MiqAeAnsibleMethodBase.ansible_stats_from_hash(@persist_state_hash)
     end
 
+    def set_service_var(name, value)
+      if service_object.nil?
+        $miq_ae_logger.error("Service object not found in root object, set_service_var skipped for #{name} = #{value}")
+        return
+      end
+
+      service_object.root_service.set_service_vars_option(name, value)
+    end
+
+    def service_var_exists?(name)
+      return false unless service_object
+
+      service_object.root_service.service_vars_options.key?(name)
+    end
+
+    def get_service_var(name)
+      return unless service_var_exists?(name)
+
+      service_object.root_service.get_service_vars_option(name)
+    end
+
+    def delete_service_var(name)
+      return unless service_var_exists?(name)
+
+      service_object.root_service.delete_service_vars_option(name)
+    end
+
     def prepend_namespace=(ns)
       @workspace.prepend_namespace = ns
     end
@@ -331,6 +358,10 @@ module MiqAeMethodService
     end
 
     private
+
+    def service_object
+      current['service'] || root['service']
+    end
 
     def editable_instance?(path)
       dom, = MiqAeEngine::MiqAePath.get_domain_ns_klass_inst(path)
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service.rb
@@ -48,6 +48,31 @@ module MiqAeMethodService
       end
     end
 
+    def service_vars_options
+      @object.options[:service_vars] ||= HashWithIndifferentAccess.new
+    end
+
+    def get_service_vars_option(key)
+      service_vars_options[key]
+    end
+
+    def set_service_vars_option(key, value)
+      ar_method do
+        service_vars_options[key] = value
+        @object.update(:options => @object.options)
+      end
+    end
+
+    def delete_service_vars_option(key)
+      return unless service_vars_options&.key?(key)
+
+      ar_method do
+        service_vars_options.delete(key).tap do
+          @object.update(:options => @object.options)
+        end
+      end
+    end
+
     def name=(new_name)
       ar_method do
         @object.update!(:name => new_name)
--- a/spec/miq_ae_service_spec.rb
+++ b/spec/miq_ae_service_spec.rb
@@ -134,6 +134,63 @@ describe MiqAeMethodService::MiqAeService do
       expect(miq_ae_service.ansible_stats_vars).to eq('var1' => 'value1', 'var2' => 'value2')
     end
   end
+
+  context 'service_var method' do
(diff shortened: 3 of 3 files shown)


tinaafitz:
@bdunne @billfitzgerald0120 Please review.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -92,6 +92,37 @@ def ansible_stats_vars
       MiqAeEngine::MiqAeAnsibleMethodBase.ansible_stats_from_hash(@persist_state_hash)
     end
 
+    def service_object
+      current['service'] || root['service']
+    end
+
+    def set_service_var(name, value)
+      unless service_object
```
bdunne: Can this use the positive case instead?

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -48,6 +48,30 @@ def set_dialog_option(key, value)
...
+      @object.options[:service_vars] || HashWithIndifferentAccess.new
+    end
+
+    def get_service_vars_option(key)
+      service_vars_options[key]
+    end
+
+    def set_service_vars_option(key, value)
+      ar_method do
+        @object.options[:service_vars] ||= HashWithIndifferentAccess.new
```
bdunne: Do we need to use `HashWithIndifferentAccess`?  I know we've had problems deserializing after upgrades before.  cc @jrafanie

spec/miq_ae_service_spec.rb
```diff
@@ -134,6 +134,63 @@
...
+    def invoke_ae
+      MiqAeEngine.instantiate("/EVM/AUTOMATE/test1?Service::service=#{@service.id}", user)
+    end
+
+    before do
+      Spec::Support::MiqAutomateHelper.create_service_model_method('SPEC_DOMAIN', 'EVM', 'AUTOMATE', 'test1', 'test')
+      @ae_method = ::MiqAeMethod.first
+
+      @parent = FactoryBot.create(:service)
+      @service = FactoryBot.create(:service, :name => "test_service", :description => "test_description", :service => @parent)
```
bdunne: Please don't set `:name` as it is sequenced by the factory.

bdunne: Also, is `:description` needed?

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -48,6 +48,30 @@ def set_dialog_option(key, value)
...
+        @object.options[:service_vars][key] = value
+        @object.update(:options => @object.options)
+      end
+    end
+
+    def delete_service_vars_option(key)
+      ar_method do
+        @object.options[:service_vars]&.delete(key).tap do |value|
+          @object.update(:options => @object.options) if value
```
bdunne: So, we don't want to save the delete of the key if the value is `nil` or `false`?

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
bdunne: IIRC, the namespace of HWIA changed in a Rails version upgrade and we had to create an alias or something in order to deserialize them, but I can't find it right now so maybe I'm thinking of something else.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -48,6 +48,32 @@ def set_dialog_option(key, value)
       end
     end
 
+    def service_vars_options
+      @object.options[:service_vars] || HashWithIndifferentAccess.new
```
bdunne: Should this be `||=`?

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -48,6 +48,32 @@ def set_dialog_option(key, value)
...
+    end
+
+    def get_service_vars_option(key)
+      service_vars_options[key]
+    end
+
+    def set_service_vars_option(key, value)
+      ar_method do
+        @object.options[:service_vars] ||= HashWithIndifferentAccess.new
```
bdunne: Can the `service_vars_options` method be reused here rather than duplicating it?

lfu: This is the ```set``` method. The other one is the ```get``` method to retrieve data from DB.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
lfu: I don't think so.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -92,6 +92,37 @@ def ansible_stats_vars
       MiqAeEngine::MiqAeAnsibleMethodBase.ansible_stats_from_hash(@persist_state_hash)
     end
 
+    def service_object
```
gmcculloug: Is this only meant for internal use?  If so it can be moved to the `private` section of the file.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
gmcculloug: I think @bdunne is correct that this should reuse the `service_vars_options` method and that the `service_vars_options` method should use `||=` as mentioned above.  That would reduce the places that have to know the options hash key and the `HashWithIndifferentAccess` logic.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -48,6 +48,32 @@ def set_dialog_option(key, value)
...
+        @object.options[:service_vars][key] = value
+        @object.update(:options => @object.options)
+      end
+    end
+
+    def delete_service_vars_option(key)
+      return unless @object.options[:service_vars]&.key?(key)
```
gmcculloug: This could use `service_vars_options` as well.

lfu:
@miq-bot add_label changelog/yes

PR #339, opened 2019-07-16
Don't use string interpolation inside gettext

mzazrivec:
We must not use string interpolation with `#{}` in gettext strings: http://manageiq.org/docs/guides/i18n#string-interpolation

Strings that come out of string interpolation with `#{}` are in fact new strings, which are not present in the catalog. We must use `%{}` instead.

@pkomanek @tinaafitz

### DIFF
--- a/app/models/miq_ae_yaml_import.rb
+++ b/app/models/miq_ae_yaml_import.rb
@@ -274,18 +274,23 @@ class MiqAeYamlImport
   def playbook_attr_error_msg(attr, options)
     case attr
     when 'repository'
-      _("Repository '#{options[:repository_name]}' not found in database. Please try and import this repo "\
-        "into this appliance and retry the import. If the repository has been deleted this import will never succeed.")
+      _("Repository '%{repository_name}' not found in database. Please try and import this repo "\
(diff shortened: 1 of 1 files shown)


tinaafitz:
Hi @mzazrivec Thanks for bringing that to my attention. I appreciate it. :-)

I see that the options hash reference was removed in a few places, for example,
```'#{options[:repository_name]}'``` was changed to ```'%{repository_name}'``` and was wondering why that is.

**Comment by mzazrivec** [comment]
@tinaafitz 

```ruby 
whatever = rand(10)
_("abc #{whatever} def")
```

means that the string `"abc #{whatever} def"` will be evaluated first, then gettext will be called. And since the string inside gettext will be always dynamic, gettext will not know how to translate the string.

```ruby
whatever = rand(10)
_("abc %{whatever} def") % {:whatever => whatever}
```

means that first gettext is called, it will know how to translate the string inside (since it's a static string) and the string expansion takes place afterwards.

tinaafitz:
@mzazrivec Awesome, thanks so much for the explanation.

mzazrivec:
Can this be merged please? Thanks.

PR #342, opened 2019-07-26
[IVANCHUK] Change travis.yml to use the only supported ruby, 2.5.3

d-m-u:
Change travis.yml to use the only supported ruby, 2.5.3

Related to ManageIQ/manageiq#19050

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
+---
 dist: xenial
 language: ruby
 rvm:
-- 2.4.6
 - 2.5.3
 sudo: false
 cache: bundler


gmcculloug:
Is there a related PR in the main repo [ManageIQ/manageiq](https://github.com/ManageIQ/manageiq) that we can reference in the description?  I'd like to make sure we are making these changes as part of a coordinated effort.

jrafanie:
the main repo PR was this: https://github.com/ManageIQ/manageiq/pull/19050

gmcculloug:
Thanks @jrafanie.  I'd recommend adding that to the description when we create these on all the repos.  It is a useful reference and also helpful to see that it is already merged.

PR #345, opened 2019-08-05
Add method $evm.get_state_vars

lfu:
New $evm.get_state_vars method to get all state_vars without having to know the names. 

https://bugzilla.redhat.com/show_bug.cgi?id=1734636

@miq-bot add_label enhancement, changelog/yes, ivanchuk/no

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
@@ -88,6 +88,10 @@ module MiqAeMethodService
       @persist_state_hash.delete(name)
     end
 
+    def get_state_vars
+      @persist_state_hash
+    end
+
     def ansible_stats_vars
       MiqAeEngine::MiqAeAnsibleMethodBase.ansible_stats_from_hash(@persist_state_hash)
     end
--- a/spec/miq_ae_service_spec.rb
+++ b/spec/miq_ae_service_spec.rb
@@ -127,6 +127,13 @@ describe MiqAeMethodService::MiqAeService do
       end
     end
 
+    it 'get_state_vars' do
+      allow(workspace).to(receive(:disable_rbac))
+      expect(miq_ae_service.get_state_vars).to eq({})
+      miq_ae_service.instance_eval { @persist_state_hash = { 'var1' => 'value1', 'var2' => 'value2' } }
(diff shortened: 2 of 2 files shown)


mkanoor:
@lfu we have internal state_vars that we use for retry, should we exclude them from this list. There is also a YAML string payload for state machine info.

mkanoor:
@lfu In the state vars we have system vars, user vars and ansible vars all clubbed into one. So exposing all of them is going to be a problem. For now can we limit this to just ansible vars and do the user vars in a separate one. We also need to prevent users from updating system state vars and messing up the state machine processing.

https://github.com/ManageIQ/manageiq-automation_engine/blob/0c498633308578ba84bfb52328234db1ece7c659/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb#L75

We would have to guard the system vars from getting edited by automated methods
The current list of system vars are here
https://github.com/ManageIQ/manageiq-automation_engine/blob/0c498633308578ba84bfb52328234db1ece7c659/lib/miq_automation_engine/engine/miq_ae_engine.rb#L59
Some of the values are YAML strings

tinaafitz:
@lfu Initially I thought we should include all state_vars, but after discussing it with Madhu, I think we should limit it. I know we already have this: https://github.com/ManageIQ/manageiq-automation_engine/pull/336

mkanoor:
@lfu @tinaafitz 
My bad. Please ignore the previous comment we are only dumping out the ae_state_data
https://github.com/ManageIQ/manageiq-automation_engine/blob/0c498633308578ba84bfb52328234db1ece7c659/lib/miq_automation_engine/engine/miq_ae_engine.rb#L66
So the other fields are ok.
The only thing I would consider is that the the ansible state vars are in a separate key so the caller would have to access that key to get to the variables.

mkanoor:
@lfu 
All of the user defined variables are going to be prefixed with
ansible_stats_

https://github.com/ManageIQ/manageiq-automation_engine/blob/0c498633308578ba84bfb52328234db1ece7c659/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_ansible_method_base.rb#L5

Is that a concern? 
Since this is targeted only to Ruby Automate methods it might be OK, the ruby method can add new state_vars prefixed with ansible_stats and it will flow to the next playbook if any.

PR #349, opened 2019-08-15
Undefine Kernel#display which would be called over service#display for DRbObject

lfu:
Display is a column in MIQ services table. But Kernel has a method named as display. 

When running an automation inline method, ```service.display``` would call ```Kernel#display``` instead of getting the value from the table column.

https://bugzilla.redhat.com/show_bug.cgi?id=1737559

@miq-bot add_label bug, hammer/yes, Ivanchuk/yes, changelog/yes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -117,6 +117,8 @@ begin
 
   DRbObject.send(:undef_method, :inspect)
   DRbObject.send(:undef_method, :id) if DRbObject.respond_to?(:id)
+  # undefine Object#display which would be called over service#display
+  DRbObject.send(:undef_method, :display)
 
   DRb.start_service
   $evmdrb = DRbObject.new_with_uri(MIQ_URI)


gmcculloug:
Since we know this variable name is problematic does it make sense to create an alias at the active_record level.

For example:
```ruby
  alias_attribute :display_in_catalog, :display
  virtual_column :display_in_catalog, :type => :boolean
```
This would give us a path to renaming it.
There is also a `deprecate_attribute_methods` that we might be able to utilize.

lfu:
> ```ruby
>   virtual_column :display_in_catalog, :type => :boolean
> ```
What is the purpose of this line?

gmcculloug:
@lfu The `virtual_column` line allows the new alias attribute name to flow through the Automate service model.

simaishi:
Hammer backport details:

```
$ git log -1
commit 295c9ca9ff645541884217765b7d5510a98ffed7
Author: tina <tfitzger@redhat.com>
Date:   Mon Aug 19 13:51:57 2019 -0400

    Merge pull request #349 from lfu/service_display_1737559
    
    Undefine Kernel#display which would be called over service#display for DRbObject
    
    (cherry picked from commit 3cf59e26a94f8284fc8cc1f4183c0f6360aa10c0)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1745763
```

PR #352, opened 2019-08-20
Expose encode_with to generic object.

lfu:
This method is called by YAML.dump when miq_queue sets the callback result into miq_task.task_results.

Reported by https://github.com/ManageIQ/manageiq-ui-classic/pull/6040#issuecomment-522771465.

@miq-bot add_label bug, Ivanchuk/yes

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_generic_object.rb
@@ -18,6 +18,11 @@ module MiqAeMethodService
       ar_method { wrap_results(@object.remove_from_service(Service.find_by(:id => service.id))) }
     end
 
+    # This method is called by YAML.dump when miq_task#queue_callback sets the result into miq_task.task_results.
+    def encode_with(*args)
+      ar_method { wrap_results(@object.encode_with(*args)) }
+    end
+
     private
 
     def ae_user_identity


martinpovolny:
This should be the last fix needed to resolve 
https://bugzilla.redhat.com/show_bug.cgi?id=1550002

gmcculloug:
@martinpovolny I was talking to @lfu about looking into this issue deeper as the `MiqAeEngine::<class>` objects should only be referenced by the automate engine, they should not end up on the queue like we are seeing in the logging.

`[----] I, [2019-08-16T21:12:06.499520 #28794:1116600]  INFO -- : MIQ(MiqQueue#m_callback) Message id: [55022599], Invoking Callback with args: ["Finished", "ok", "Message delivered successfully", "#<MiqAeEngine::MiqAeWorkspaceRuntime:0x000000000ab47350 @readonly=false, @nodes=[#<MiqAeEngine::MiqAeObject:0x000000000985ec20 @workspace=#<MiqAeEngine::MiqAeWorkspaceRuntime:0x000000000ab47350 ...>, @namespace=\"ManageIQ/SYSTEM\", @klass=\"PROCESS\", @instance=\"Request\", @attributes={\"object_name\"=>\"Request\", \"request\"=>\"url_open_go\", \"vmdb_object_type\"=>\"generic_object\", \"generic_object_id\"=>\"12\", \"generic_object\"=>#<MiqAeServiceGenericObject:0x4294614 @object=#<GenericObject:0x00000000085293f8 id..."]`

Do you know where this callback is being setup?

lfu:
@gmcculloug @mkanoor 
[custom button](https://github.com/ManageIQ/manageiq/blob/5f85fc0b39ae1b54da1153d1aaead7bce67f394d/app/models/custom_button.rb#L149) calls ```MiqTask.generic_action_with_callback``` which sets up the callback. 
And ```MiqAeEngine.deliver``` always [return a workspace handle](https://github.com/ManageIQ/manageiq-automation_engine/blob/69b58ee4ce23eef1591a90c145e8b225821a80b9/lib/miq_automation_engine/engine/miq_ae_engine.rb#L124) which references the generic object in this case after calling the automate method.

mkanoor:
@lfu @gmcculloug I am not sure if saving the whole workspace is the correct way of doing this, in the past the UI has been running dynamic dialog via automate and it would run it synchronously and use the specific attributes from the workspace and discard the workspace. This runs it  asynchronously which requires the callback and then serializing the whole workspace to pick a single attribute to fetch the URL. Doing it synchronously like the UI invokes Dynamic Dialogs might feel like a better approach. Or if we want to run it asynchronously we should have a special entry point in Automate that can give you back a YAML/XML payload tree of objects without the whole serialized workspace. That approach is being used when we display the results of simulation from the UI which also runs it synchronously.

mkanoor:
@lfu This is how the UI simulate does it
    https://github.com/ManageIQ/manageiq-ui-classic/blob/24c86b11b6afc58c96a35cfecffd100ebc2a67c2/app/controllers/application_controller/automate.rb#L143
Please note how it calls 
@results = ws.to_expanded_xml

Which converts the workspace to a plain XML payload which the UI can then render.

martinpovolny:
The UI is not calling the automate directly in this case. It's queueing a task. Automate code is being called from a worker.

gmcculloug:
My understanding is that the UI is not using the workspace at all, it just wants to wait until the automation work is done and then see if the object has an "external url" defined.  The issue is that the `MiqTask.generic_action_with_callback` captures the output of the method `MiqAeEngine.deliver` which is returning the AE workspace.  (As @lfu points out above https://github.com/ManageIQ/manageiq-automation_engine/pull/352#issuecomment-523589244)

martinpovolny:
> My understanding is that the UI is not using the workspace at all, it just wants to wait until the automation work is done and then see if the object has an "external url" defined. 

That is correct. At least for this case. I do not understand the internals enough to say if that is the case for all invocations of Automate throught a custom button. (Checking the external_url is just one of the cases.)

> The issue is that the MiqTask.generic_action_with_callback captures the output of the method MiqAeEngine.deliver which is returning the AE workspace. (As @lfu points out above #352 (comment))

That it past my current lever of understanding of the matter.

mkanoor:
@lfu A better approach is to pass extra arguments into the Automate Engine which says that you want the result to be in a specific format or pick a specific attribute from the object

e.g
'result_format' => 'json'
'result_path'    => 'root.external_url'

Then when we return at the end of the engine execution we can convert the data to the appropriate format. The workspace has a method called hash_workspace which will give you the data as a hash which can be converted to json or yaml.

Here is where we do the output format analysis
https://github.com/ManageIQ/manageiq-automation_engine/blob/db1e69699724815ea7f2fc593e306f1a2de68815/lib/miq_automation_engine/engine/miq_ae_engine.rb#L124

The custom button UI code would have to tweaked to pass in the new result format attributes and we can streamline what we store in the task results. This same approach could be used by the Dynamic Dialogs if they want to run the methods async.

lfu:
@martinpovolny 
UI code needs to add ```:result_format => 'ignore'``` when creating the custom button and adding ```"request" => "test_open_url"```.

martinpovolny:
> UI code needs to add :result_format => 'ignore' when creating the custom button and adding "request" => "test_open_url". I will work on the backend to support that option.

I am sorry, but that is rather vague for me to turn into a PR.

What call should add the first key/value pair? And what call should add the second key/value pair?

lfu:
> What call should add the first key/value pair? And what call should add the second key/value pair?

The place where the external_url is got saved into VMDB, please add ```:result_format => 'ignore' ``` into the same hash.

martinpovolny:
> The place where the external_url is got saved into VMDB, please add :result_format => 'ignore' into the same hash.

That is still a riddle to me. What hash?

Do you mean here?
https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/mixins/miq_ae_external_url_mixin.rb

Or here?
https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/external_url_mixin.rb#L10

lfu:
Screen Shot 2019-08-22 at 14 34 59

The place where ae_attributes is populated.

martinpovolny:
> `ae_attributes` is populated

ok, so you are speaking about some entity called `resource_action` that has an atribute `ae_attributes` and some of the records there are related to a custom button, the one that is being pressed.

And at some point there's a value being added there (the one related to my custom button) and you want me to populate that column with value 'ignore'.

That has little to do with `The place where the external_url is got saved into VMDB` is that correct?

lfu:
@martinpovolny Never mind. No change to UI. I will do it from backend. Sorry for the confusion.

gmcculloug:
@lfu Can we close this PR and open a new one for the changes you are working on?

lfu:
Colse to favor https://github.com/ManageIQ/manageiq-automation_engine/pull/354.

PR #354, opened 2019-08-23
Add MiqAeEngine.return_result.

lfu:
Returning back workspace breaks the custom button for generic objects.

Add new options that can be passed in as ```args``` to ```MiqAeEngine.deliver```.
- result_format 
       ```ignore``` - To stop automate send back workspace object as return value
       ```nil```        - Default value, workspace object would be returned by ```MiqAeEngine.deliver```
- result_on_success
User can use this option to set the expected return value from ```MiqAeEngine.deliver```. 
Used together with ```result_format```.

Blocks https://github.com/ManageIQ/manageiq/pull/19195.

Replace https://github.com/ManageIQ/manageiq-automation_engine/pull/352.

Reported by https://github.com/ManageIQ/manageiq-ui-classic/pull/6040#issuecomment-522771465.
https://bugzilla.redhat.com/show_bug.cgi?id=1550002

@miq-bot add_label bug, Ivanchuk/yes, changelog/yes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -121,7 +121,7 @@ module MiqAeEngine
         end
       end
 
-      return ws
+      return_result(ws, options[:attrs])
     rescue MiqAeException::Error => err
       message = "Error delivering #{automate_attrs.inspect} for object [#{object_name}] with state [#{state}] to Automate: #{err.message}"
       miq_task&.error(MiqTask::MESSAGE_TASK_COMPLETED_UNSUCCESSFULLY)
@@ -132,6 +132,13 @@ module MiqAeEngine
         miq_task.update_message(MiqTask::MESSAGE_TASK_COMPLETED_SUCCESSFULLY) if miq_task.message == MiqTask::DEFAULT_MESSAGE
         miq_task.state_finished
       end
+    end
+  end
+
+  def self.return_result(workspace, options)
+    case options["result_format"]
+    when 'ignore' then options["result_on_success"] || 'Ok'
+    when nil then workspace
     end
   end
 
--- a/spec/miq_ae_engine_spec.rb
+++ b/spec/miq_ae_engine_spec.rb
@@ -793,6 +793,20 @@ describe MiqAeEngine do
     expect(my_objects_array.length).to eq(0)
     expect(my_objects_array).to eq([])
   end
+
+  context ".return_result" do
+    let(:workspace) { double('MiqAeEngine::MiqAeWorkspaceRuntime') }
+
+    it "returns workspace if result_format=nil" do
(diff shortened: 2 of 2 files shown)


mkanoor:
@lfu can you put the description of the keywords in the PR

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -135,6 +135,13 @@ def self.deliver(*args)
     end
   end
 
+  def self.return_result(workspace, options)
+    case options["result_format"]
+    when 'ignore' then options["result_on_success"] || 'Ok'
+    when nil then workspace
```
gmcculloug: Should `workspace` just be the `else` case?

Are there current plans to use `options["result_on_success"]`?

mkanoor: @gmcculloug Since the callback is not looking at the result I am not sure how it figures out if there was an error. We thought this attribute allows the caller to ignore the result but return a precanned value on success which can be checked in the callback.

gmcculloug: Thanks @mkanoor I'm not against it, just wasn't sure if you had plans to use it with the custom button work.

For my `else` comment above I realized keeping it in the current form makes it obvious that if they user passes a `result_format` they need to specific a supported type or it will return `nil`.  I think that is reasonable.  👍

himdel:
I'm currently unable to use an inline method to provide a default value for a service dialog dynamic field.

The method is just:

```
$evm.root['default_value'] = "1/2/3"
exit MIQ_OK
```

During the `/api/service_dialogs/114?resource_action_id=7174&target_id=93&target_type=service_template` api request, I'm getting a 500 HTTP error - `invalid date`. That's because in `DialogFieldDateControl#value`, `@value` is `<Script error>`.

=> And the exception that lead to that error (from the rescue in `DynamicDialogFieldValueProcessor`) is `undefined method `root' for \"Ok\":String` back.

When I change `return_result` here to just return `workspace` it works.

But, `options` is `{"dialog_text_box_1"=>"", "dialog_date_control_1"=>nil, "dialog_date_time_control_1"=>nil, "result_format"=>"ignore"}`, so theoretically this should be happening.

Any idea what could be wrong please?
(I was just ordering a generic catalog item as usual to get to testing a service dialog..)

mkanoor:
@himdel 
There is a set format for sending results back to a dynamic dialog, it needs to have some preset keys defined.

You can look at this method for an example
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb

I think this one lists the available_projects in Openshift.

himdel:
Yup, my issue is that `result_format` is always set to `ignore` when running the service dialog -  https://github.com/ManageIQ/manageiq/pull/19407/files#diff-2cf77def0b1215bf4a0103936e517e2fL33

PR #358, opened 2019-09-06
Add method to delete tag

astrozzc:
Add method to delete tag
    
Both tag_delete and tag_delete! are added.
    
For tag_delete!
1. When tag is found and not assigned, it would be deleted and return the entry;
2. When tag is assigned, user will be warned and tag could not be deleted;
3. When entry is not found, user will be warned, in case of e.g. typo;
    
tag_delete behaves similar to tag_delete! except that it returns true when
execution succeeds and will suppress all exceptions and return false.
    
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1744514#c4

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
@@ -99,6 +99,27 @@ module MiqAeMethodService
       end
     end
 
+    def self.tag_delete!(category, entry)
+      ar_method do
+        cat = Classification.find_by_name(category)
+        raise "Category <#{category}> does not exist" if cat.nil?
+
+        ent = cat.find_entry_by_name(entry)
+        raise "Entry <#{entry}> does not exist" if ent.nil?
+
+        raise "This tag has assignments. Please delete assignments before deleting the tag." if AssignmentMixin.all_assignments(ent.tag.name).present?
+
+        ent.destroy!
+      end
+    end
+
+    def self.tag_delete(category, entry)
+      tag_delete!(category, entry)
+      true
+    rescue StandardError
+      false
+    end
+
     def self.create_provision_request(*args)
       # Need to add the username into the array of params
       # TODO: This code should pass a real username, similar to how the web-service
--- a/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
@@ -163,6 +163,80 @@ describe MiqAeMethodService::MiqAeServiceMethods do
     expect(ct.entries.collect(&:name).include?('fred')).to be_truthy
   end
 
+  context "#tag_delete!" do
+    let(:ct) { FactoryBot.create(:classification_department_with_tags) }
+    let(:entry_name) { ct.entries.first.name }
+
+    it "could delete tag if it is not assigned" do
+      tag_name = "/managed/#{ct.name}/#{entry_name}"
+      expect(Tag.exists?(:name => tag_name)).to be_truthy
+      method = "$evm.root['#{@ae_result_key}'] = $evm.execute(:tag_delete!, #{ct.name.inspect}, #{entry_name.inspect})"
+      @ae_method.update(:data => method)
+
+      expect(invoke_ae.root(@ae_result_key)).to be_truthy
+      expect(Tag.exists?(:name => tag_name)).to be_falsey
+    end
+
+    it "could not delete tag if it is assigned" do
+      assignment_tag = "/chargeback_rate/assigned_to/vm/tag/managed/#{ct.name}/#{entry_name}"
+      Tag.create!(:name => assignment_tag)
+      expect(Tag.exists?(:name => assignment_tag)).to be_truthy
+
+      method = "$evm.root['#{@ae_result_key}'] = $evm.execute(:tag_delete!, #{ct.name.inspect}, #{entry_name.inspect})"
+      @ae_method.update(:data => method)
+
+      expect { invoke_ae.root(@ae_result_key) }.to raise_error(MiqAeException::UnknownMethodRc)
+      expect(Tag.exists?(:name => assignment_tag)).to be_truthy
+    end
+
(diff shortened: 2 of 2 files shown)


astrozzc:
@miq-bot add_label enhancement, ivanchuk/no

astrozzc:
@miq-bot assign @tinaafitz @mkanoor

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -99,6 +99,23 @@ def self.tag_create(category, options = {})
       end
     end
 
+    def self.tag_delete(category, entry, force_delete = false)
+      ar_method do
+        cat = Classification.find_by_name(category)
```
gmcculloug: Note, this `find_by_name` call is ok since the model is overriding the method: https://github.com/ManageIQ/manageiq/blob/master/app/models/classification.rb#L338

astrozzc: Oh, thanks for pointing this out! I was trying to fix this but failed!

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -99,6 +99,23 @@ def self.tag_create(category, options = {})
       end
     end
 
+    def self.tag_delete(category, entry, force_delete = false)
```
gmcculloug: My suggestion for the `force_delete` flag was a thought on what we might **consider in the future** if we find there is enough need for it.  I suggest we do not expose that at this time and wait until there is a clear use case for it.

spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
```diff
@@ -163,6 +163,53 @@ def category_create_script
...
   end
 
+  context "#tag_delete" do
+    before(:each) do
+      @ct = FactoryBot.create(:classification_department_with_tags)
+      @ct_name = @ct.name
+      @entry_name = @ct.entries.first.name
+      @assignment_tag = "/chargeback_rate/assigned_to/vm/tag/managed/#{@ct_name}/#{@entry_name}"
+      Tag.create!(:name => @assignment_tag)
+    end
```
gmcculloug: The `before(:each)` is an older syntax and should just be `before` now.

I think the whole section would could be done a little differently:

```ruby
context "#tag_delete" do
  let(:ct) { FactoryBot.create(:classification_department_with_tags) }
  let(:entry_name) { ct.entries.first.name }
  let(:assignment_tag) { "/chargeback_rate/assigned_to/vm/tag/managed/#{ct.name}/#{entry_name}" }
  before { Tag.create!(:name => assignment_tag) }

  ...
```

I removed the `ct_name` variable as I don't think it is really needed.  Changing to `let` blocks also means the variable references in the tests no longer require the `@` reference.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
astrozzc: Sounds good to me. Since it is not hard to add this flag, I will remove it from this PR now and wait for more use cases.

spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb
astrozzc: Got it! let looks much better

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -99,6 +99,21 @@ def self.tag_create(category, options = {})
...
+        cat = Classification.find_by_name(category)
+        raise "Category <#{category}> does not exist" if cat.nil?
+
+        ent = cat.find_entry_by_name(entry)
+        raise "Entry <#{entry}> does not exist" if ent.nil?
+
+        raise "This tag has assignments. Please delete assignments before deleting the tag." if AssignmentMixin.all_assignments(ent.tag.name).present?
+
+        ent.destroy
+        true
```
mkanoor: @astrozzc Is this true needed? Is it there to suppress the ent.destroy returning a ent. The true return indicates that their could be a false return too, which are exceptions.

mkanoor: @astrozzc You might want to explore the ! methods in rails which raise an exception otherwise they might return false and suppress the exception. save!, create! etc.

astrozzc: Yes, it is to suppress the return of ent obj. I add it to make it consistent with the create method. If there is something wrong with the .destroy, then the return would be false or exception.

gmcculloug: Looks like the solution here is to use `destroy!` to ensure the method is successful or raises an error.

astrozzc: Thanks for the suggestion!

gmcculloug: Ruby allows a simpler syntax here:

```ruby
def self.tag_delete(category, entry)
  tag_delete!(category, entry)
  true
rescue
  false
end
```

gmcculloug:
@mkanoor PTAL

simaishi:
Backport approved by @gmcculloug 

Ivanchuk backport details:

```
$ git log -1
commit 8d53dd4f0a8de7718f55452df7ec575026a8dc29
Date:   Wed Sep 11 16:04:55 2019 -0400

    Merge pull request #358 from astrozzc/tag_delete
    
    Add method to delete tag
    
    (cherry picked from commit cbd08ec03e441bc051b543a7b9d4ad528aa6f76f)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1767901
```

PR #367, opened 2019-09-13
update_attributes has been deprecated

d-m-u:
```update_attributes``` was replaced by ```update``` and this needed to happen sometime

### DIFF
--- a/app/models/miq_ae_yaml_import.rb
+++ b/app/models/miq_ae_yaml_import.rb
@@ -91,7 +91,7 @@ class MiqAeYamlImport
       else
         import_all_namespaces(domain_folder, domain_obj, domain_name)
       end
-      update_attributes(domain_obj) if domain_obj
+      update(domain_obj) if domain_obj
       domain_obj
     end
   end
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

**Comment by gmcculloug** [comment]
To clarify `update_attributes` and `update_attributes!` are deprecated starting in Rails 6.
See https://rubyinrails.com/2019/04/09/rails-6-1-activerecord-deprecates-update-attributes-methods/ for details and Rails PR links.

File: app/models/miq_ae_yaml_import.rb [hunk 1]
```diff
@@ -310,10 +310,10 @@ def new_domain_name_valid?
     true
   end
 
-  def update_attributes(domain_obj)
+  def update(domain_obj)
```
gmcculloug: All these changes are pretty straightforward replacements.  Changing this method name is the only one that has potentially more exposure.

@mkanoor Do you know if this method would be called from any other repo where we might need to make the corresponding name change?

d-m-u: Oh that was me being an idiot, sorry, it was late. This shouldn't be here.

gmcculloug: Maybe/maybe not.  It is likely this name was picked to match the Rails method, so changing it to match may be the right thing.  Let's give @mkanoor  some time to respond before changing.

app/models/miq_ae_yaml_import.rb
```diff
@@ -310,10 +310,10 @@ def new_domain_name_valid?
...
     return if domain_obj.name.downcase == MiqAeDatastore::MANAGEIQ_DOMAIN.downcase
     attrs = @options.slice('enabled', 'source')
-    domain_obj.update_attributes(attrs) unless attrs.empty?
+    domain_obj.update(attrs) unless attrs.empty?
```
mkanoor: 👍

app/models/miq_ae_yaml_import.rb
gmcculloug: Confirmed that https://github.com/ManageIQ/manageiq-automation_engine/pull/367#pullrequestreview-288041339 was in response to my question here.  So we are good with the method name change.

Thanks @d-m-u

PR #370, opened 2019-09-16
Correct style using rubocop (part 4)

astrozzc:
Cops used:
Naming/FileName
Naming/ClassAndModuleCamelCase
Naming/HeredocDelimiterNaming
Naming/UncommunicativeBlockParamName

Style/AccessModifierDeclarations
Style/ClassVars
Style/MultilineIfModifier
Style/MultipleComparison
Style/MultilineWhenThen
Style/Next
Style/RaiseArgs
Style/RescueModifier
Style/RescueStandardError
Style/SafeNavigation
Style/StderrPuts

Performance/RedundantMatch
Performance/RegexpMatch

@miq-bot add_label  ivanchuk/no, refactoring, changelog/yes

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
@@ -55,7 +55,7 @@ module MiqAeEngine
 
       begin
         return MiqAeBuiltinMethod.send(mname, obj, inputs)
-      rescue => err
+      rescue StandardError => err
         raise MiqAeException::AbortInstantiation, err.message
       ensure
         # Destroy service to avoid storing object references
@@ -110,7 +110,7 @@ module MiqAeEngine
       rc, msg = run_method(*cmd)
       if ws
         ws.reload
-        ws.setters.each { |uri, value| workspace.varset(uri, value) } unless ws.setters.nil?
+        ws.setters&.each { |uri, value| workspace.varset(uri, value) }
         ws.delete
       end
       process_ruby_method_results(rc, msg)
@@ -228,8 +228,8 @@ module MiqAeEngine
         msg = "Method exited with rc=#{verbose_rc(rc)}"
         method_pid = nil
         threads = []
-      rescue => err
-        STDERR.puts "** AUTOMATE ** Method exec failed because #{err.class}:#{err.message}" if ENV.key?("CI")
+      rescue StandardError => err
+        STDERR.puts "** AUTOMATE ** Method exec failed because #{err.class}:#{err.message}" if ENV.key?("CI") # rubocop:disable Style/StderrPuts
         $miq_ae_logger.error("Method exec failed because (#{err.class}:#{err.message})")
         rc = MIQ_ABORT
         msg = "Method execution failed"
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
@@ -182,7 +182,7 @@ module MiqAeEngine
     def fetch_field_value(f)
       Benchmark.current_realtime[:fetch_field_value_count] += 1
       Benchmark.realtime_block(:fetch_field_value_time) do
-        @aei.get_field_value(f, false) unless @aei.nil?
+        @aei&.get_field_value(f, false)
       end.first
     end
 
@@ -198,16 +198,18 @@ module MiqAeEngine
       when 'Symbol'                   then xml.Symbol(value.to_s)
       when 'TrueClass', 'FalseClass'  then xml.Boolean(value.to_s)
       when /MiqAeMethodService::(.*)/ then xml.tag!($1.gsub(/::/, '-'), :object_id => value.object_id, :id => value.id)
-      when 'Array'                    then xml.Array  do
-        value.each_index do |i|
-          xml.Element(:index => i + 1) { attribute_value_to_xml(value[i], xml) }
+      when 'Array'
+        xml.Array do
+          value.each_index do |i|
+            xml.Element(:index => i + 1) { attribute_value_to_xml(value[i], xml) }
+          end
         end
-      end
-      when 'Hash'                     then xml.Hash do
-        value.each do |k, v|
-          xml.Key(:name => k.to_s) { attribute_value_to_xml(v, xml) }
+      when 'Hash'
+        xml.Hash do
+          value.each do |k, v|
+            xml.Key(:name => k.to_s) { attribute_value_to_xml(v, xml) }
+          end
         end
-      end
       when 'DRb::DRbUnknown'
         $miq_ae_logger.error("Found DRbUnknown for value: #{value.inspect} in XML: #{xml.inspect}")
         xml.String(value)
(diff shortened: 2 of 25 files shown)


astrozzc:
@miq-bot assign @tinaafitz

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -11,8 +11,10 @@ class MiqAeService
 
     attr_accessor :logger
 
+    # rubocop:disable ClassVars
```
gmcculloug: I would prefer we keep this warning enabled.  We can work on changing this to use a class instance variable in a separate PR if needed.

astrozzc: Got it, thanks for pointing this out! I thought if we changed it into '@', it would be different for each instance. It is actually will be shared between instances. Updated

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -228,8 +228,8 @@ def self.run_method(cmd)
         msg = "Method exited with rc=#{verbose_rc(rc)}"
         method_pid = nil
         threads = []
-      rescue => err
-        STDERR.puts "** AUTOMATE ** Method exec failed because #{err.class}:#{err.message}" if ENV.key?("CI")
+      rescue StandardError => err
+        warn "** AUTOMATE ** Method exec failed because #{err.class}:#{err.message}" if ENV.key?("CI")
```
gmcculloug: @mkanoor Do we want to keep this one as a `STDERR.puts`?  It was added in https://github.com/ManageIQ/manageiq/pull/11627 to help debug a sporadic Travis failure.  Not sure if it is still useful.

mkanoor: @gmcculloug Yes we need that message because we need to debug travis failures. Previously when we used to see intermittent failures we could never diagnose why a method failed.

astrozzc: @mkanoor 

So we will keep the STDERR.puts and put a # rubocop:disable Style/StderrPuts.

I am wondering Is that because we don't want the error be disabled? https://github.com/rubocop-hq/rubocop/issues/4888

astrozzc:
currently, this is failing, not due to my PR
https://travis-ci.org/ManageIQ/manageiq-automation_engine/jobs/590086208#L1875

astrozzc:
> currently, this is failing, not due to my PR
> https://travis-ci.org/ManageIQ/manageiq-automation_engine/jobs/590086208#L1875

Now there is a fix for it: https://github.com/ManageIQ/manageiq/pull/19336 Tested locally, and all tests pass.

astrozzc:
The Hakiri security issue will be resolved once this PR get merged: https://github.com/ManageIQ/manageiq/pull/19348

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
gmcculloug: @astrozzc That's not exactly what I was thinking about when I suggested reverting it, but reading more into the issue we do want to avoid the situation where `warn` could be disabled.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
gmcculloug: Agreed, I was thinking we might be able to resolve this but looks like that would definitely change the existing logic.  I don't think we need to hold up this PR for that change.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
```diff
@@ -601,7 +605,7 @@ def process_assertion(f, message, args)
         rescue SyntaxError => err
           $miq_ae_logger.error("Assertion had the following Syntax Error: '#{err.message}'")
           raise MiqAeException::AssertionFailure, "Syntax Error in Assertion: <#{assertion}>"
-        rescue Exception => err
```
gmcculloug: We are running ruby scripts here with the `eval` and do need to rescue `Exception`.  There are a few others we need to keep as well.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -171,12 +171,16 @@ def self.drb_undumped(klass)
 
     def self.ar_method
       yield
-    rescue Exception => err
+    rescue StandardError => err
```
gmcculloug: Keep `Exception`

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -332,7 +328,7 @@ def object_send(name, *params)
       ar_method do
         begin
           @object.public_send(name, *params)
-        rescue Exception
+        rescue StandardError
```
gmcculloug: Keep `Exception`

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -351,12 +347,16 @@ def self.ar_method
       # In UI Worker, query caching is enabled.  This causes problems in Automate DRb Server (e.g. reload does not refetch from SQL)
       ActiveRecord::Base.connection.clear_query_cache if ActiveRecord::Base.connection.query_cache_enabled
       yield
-    rescue Exception => err
+    rescue StandardError => err
```
gmcculloug: Keep `Exception`

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
```diff
@@ -561,10 +565,10 @@ def self.convert_value_based_on_datatype(value, datatype)
       return convert_boolean_value(value)                    if datatype == 'boolean'
       return true                                            if datatype == 'TrueClass'
       return false                                           if datatype == 'FalseClass'
-      return Time.parse(value)                               if datatype == 'time' || datatype == 'Time'
-      return value.to_sym                                    if datatype == 'symbol' || datatype == 'Symbol'
+      return Time.parse(value).getlocal                      if %w[time Time].include?(datatype)
```
gmcculloug: Been thinking about these comparisons and want to suggest using `casecmp?` include?  See https://ruby-doc.org/core-2.4.0/String.html

For example:
`'time'.casecmp?(datatype)`

Many of the comparisons here could use that.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
astrozzc: That's right! Will update it

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
astrozzc: Will update, and put a rubocop disable here.

PR #378, opened 2019-10-09
Cleanup after Ansible runner.

lfu:
Depends on https://github.com/ManageIQ/manageiq/pull/19383.

https://bugzilla.redhat.com/show_bug.cgi?id=1765682

@miq-bot add_label changelog/yes, ivanchuk/yes, hammer/no

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
@@ -1,6 +1,9 @@
 module MiqAeEngine
   class MiqAePlaybookMethod < MiqAeAnsibleMethodBase
+    include AnsiblePlaybookMixin
+
     PLAYBOOK_CLASS = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Playbook
+    STACK_CLASS = ManageIQ::Providers::EmbeddedAnsible::AutomationManager::Job
 
     private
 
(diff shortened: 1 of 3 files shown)


gmcculloug:
@mkanoor The core repo PR is merged now.  Please review.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -23,16 +29,41 @@ def execute
       running_in_state_machine? ? check_task_status(task_id) : wait_for_method(task_id)
     end
 
+    def process_result(task = nil)
+      stack_job(task).refresh_ems
+      log_stdout(task)
```
gmcculloug: Two things:

1. Would `process_result` ever get called with a `nil` task value?  The `.refresh_ems` call will fail since `stack_job(task)` is likely to return nil (unless `stack_job` was previously called and just returning the cached object.)

2. Is there a reason to pass the `task` object to `log_stdout` stead of the `stack_job` object?  (The `task` object is only used to load the `stack_job` as part of the logging.)

If `task` could be `nil` them maybe something like this:
```ruby
def process_result(task = nil)
  job = stack_job(task)
  if job
    job(task).refresh_ems
    log_stdout(job.reload)
  end

  super
end
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
```diff
@@ -23,16 +29,41 @@ def execute
...
+    end
+
     def resolved_hosts
       @ae_object.substitute_value(@aem.options[:hosts], nil, true).tap do |value|
         raise ArgumentError, "Hosts field #{@aem.options[:hosts]} resolved to empty string" if value.blank?
       end
     end
+
+    def log_stdout(task)
+      playbook_log_stdout(@aem.options[:log_output], stack_job(task).reload)
```
gmcculloug: Per my other comment, I think the `reload` should be moved to the caller and not done here in a logging method.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_playbook_method.rb
lfu: `task` should always be there.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 77be7f1996c85cfc3d2c46f9fb08532aa1c26db7
Date:   Fri Oct 25 16:14:19 2019 -0400

    Merge pull request #378 from lfu/ansible_runner_cleanup

    Cleanup after Ansible runner.

    (cherry picked from commit dbee6b32a8160b3d5f729144553d6d11cab31062)

    https://bugzilla.redhat.com/show_bug.cgi?id=1784103
```

PR #379, opened 2019-10-18
Remove #retire_service_resources

bdunne:
Related to https://github.com/ManageIQ/manageiq/pull/19413

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service.rb
@@ -9,7 +9,6 @@ module MiqAeMethodService
     require_relative "mixins/miq_ae_external_url_mixin"
     include MiqAeExternalUrlMixin
 
-    expose :retire_service_resources
     expose :automate_retirement_entrypoint
     expose :start
     expose :stop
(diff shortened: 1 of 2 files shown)


gmcculloug:
This method is called from the content repo.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -9,7 +9,6 @@ class MiqAeServiceService < MiqAeServiceModelBase
     require_relative "mixins/miq_ae_external_url_mixin"
     include MiqAeExternalUrlMixin
 
-    expose :retire_service_resources
```
gmcculloug: We need to keep this `expose` as the backend method is called from content here: https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/retire_service.rb#L5

This change will turn it into a no-op which is also incorrect.  The required logic was restored when the delete was reverted in https://github.com/ManageIQ/manageiq/pull/19417.

This PR should be closed.

bdunne: Wasn't it already a no-op?

gmcculloug: Yes, you are right.  Let me review all this with @tinaafitz today.

lib/miq_automation_engine/service_models/miq_ae_service_service.rb
```diff
@@ -32,6 +31,9 @@ def self.create(options = {})
       ar_method { MiqAeServiceModelBase.wrap_results(Service.create!(attributes)) }
     end
 
+    def retire_service_resources
+    end
```
gmcculloug: @bdunne Can you just add a comment that this is for backwards compatibility with Automate scripts.  Maybe include a link to https://github.com/ManageIQ/manageiq-content/pull/590.

PR #380, opened 2019-10-21
[WIP] Override miq_ae_domain#update_attributes method.

lfu:
Per https://github.com/ManageIQ/manageiq-api/pull/647#issuecomment-543188469. 

@miq-bot add_label hammer/no, ivanchuk/no, changelog/yes, enhancement

### DIFF
--- a/app/models/miq_ae_domain.rb
+++ b/app/models/miq_ae_domain.rb
@@ -100,6 +100,17 @@ class MiqAeDomain < MiqAeNamespace
 
   alias editable_contents? editable?
 
+  def update_attributes_api(hash)
+    non_editable = hash.keys.reject { |p| editable_property?(p) }
+    if non_editable.present?
+      msg = "Non editable attributes [#{non_editable.join(",")}] in [#{hash}] for domain update"
+      $miq_ae_logger.error(msg)
+      raise MiqAeException::InvalidAttribute, msg
+    end
+
+    update!(hash)
+  end
+
   def version
     version_field = about_class.try(:ae_fields).try(:detect) { |fld| fld.name == 'version' }
     version_field.try(:default_value)
--- a/lib/miq_automation_engine/miq_ae_exception.rb
+++ b/lib/miq_automation_engine/miq_ae_exception.rb
@@ -30,6 +30,7 @@ module MiqAeException
   class ClassNotFound < MiqAeDatastoreError; end
   class InstanceNotFound < MiqAeDatastoreError; end
   class FieldNotFound < MiqAeDatastoreError; end
+  class InvalidAttribute < MiqAeDatastoreError; end
   class InvalidClass < MiqAeDatastoreError; end
   class InvalidDomain < MiqAeDatastoreError; end
   class DirectoryNotFound < MiqAeDatastoreError; end
--- a/spec/models/miq_ae_domain_spec.rb
+++ b/spec/models/miq_ae_domain_spec.rb
@@ -269,6 +269,38 @@ describe MiqAeDomain do
     end
   end
 
+  context "udpate attributes via api" do
+    let(:value1)  { {'priority' => 30, 'enabled' => true} }
+    let(:value2)  { {'name' => 'new_domain', 'description' => 'new system domain', 'priority' => 30, 'enabled' => true} }
+    it "does not update system domain" do
+      dom = FactoryBot.create(:miq_ae_system_domain, :tenant => @user.current_tenant)
+      expect { dom.update_attributes_api(value1) }.to raise_error(/Non editable attributes/)
+    end
+
+    it 'updates user domain' do
+      dom = FactoryBot.create(:miq_ae_domain, :tenant => @user.current_tenant)
+      dom.update_attributes_api(value2)
+      expect(dom.name).to eq(value2['name'])
+      expect(dom.description).to eq(value2['description'])
+      expect(dom.priority).to eq(30)
+      expect(dom.enabled).to be_truthy
+    end
+
+    context "git domain" do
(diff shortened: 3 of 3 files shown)


app/models/miq_ae_domain.rb
```diff
@@ -100,6 +100,17 @@ def editable_property?(property)
 
   alias editable_contents? editable?
 
+  def update_attributes(hash)
```
gmcculloug: To follow Rails conventions the `update_attributes` method should only return `true` or `false` and an `update_attributes!` method would be used to raise errors.  And if possible can this be changed to use `update` and `update!`

```ruby
def update(hash)
  update!(hash)
rescue
  false
end

def update!(hash)
  not_editable = hash.keys.reject { |p| editable_property?(p) }
  ...
end

```

app/models/miq_ae_domain.rb
```diff
@@ -100,6 +100,17 @@ def editable_property?(property)
...
+    not_editable = hash.keys.reject { |p| editable_property?(p) }
```
gmcculloug: Suggest using `non_editable` for the variable name and `"Non  editable attributes..."` in the message text.

app/models/miq_ae_domain.rb
```diff
@@ -100,6 +100,17 @@ def editable_property?(property)
...
+    if not_editable.present?
+      msg = "Not editable attributes [#{not_editable.join(",")}] in [#{hash}] for domain update"
+      $miq_ae_logger.error(msg)
+      raise msg
```
gmcculloug: @mkanoor Is there an existing Automate error type we should use here or go with `ActiveRecord::RecordInvalid` which the `update!` method uses.

app/models/miq_ae_domain.rb
mkanoor: @lfu I think non_editable is better than not_editable

app/models/miq_ae_domain.rb
mkanoor: @gmcculloug @lfu We don't have any exceptions for invalid attributes, the ActiveRecord::RecordInvalid is a good option to start with.

app/models/miq_ae_domain.rb
gmcculloug: @lfu @mkanoor Any thoughts on this change?  The change to use `update` depends on what the caller is using but we should be switching over to `update` anyway.

lfu:
For a git domain, this spec test expects ```git_domain.update``` can update all attributes.
https://github.com/ManageIQ/manageiq-automation_engine/blob/9ab733b5fce3e6c1361482702558d355727ad5a3/spec/models/miq_ae_domain_spec.rb#L336

But this spec test expects only `priority` and `enabled` can be updated.
https://github.com/ManageIQ/manageiq-automation_engine/blob/9ab733b5fce3e6c1361482702558d355727ad5a3/spec/models/miq_ae_domain_spec.rb#L257

We need two methods if both behavior are expected.

Since Rails 4, `update_attributes` has been an alias of `update`.
I think it is confusing if we override `update_attributes` and make it behave differently from `update`.
Can we name the new method differently and leave Rails `update/update_attributes` untouched?

mkanoor:
@lfu I think we might need a way to govern if the update is happening internally or externally. Either via a different method on the MiqAeDomain which we can call once we have updated a git domain. The values like commit_sha for a git domain will not be provided remotely by a user, it is part of the import process. This new method should allow the internals of the import process to update the following set of attributes

```
    t.string   "commit_sha"
     t.datetime "commit_time"
    t.text     "commit_message"
    t.bigint   "git_repository_id"
    t.string   "ref"
    t.string   "ref_type"
    t.datetime "last_import_on"
```

mkanoor:
@lfu I am also thinking since our goal is to block the API, and the UI sitting behind the API from updating specific attributes should we have a method called
update_attributes_api(hash)

Which the API controller can call and update specific attribute but leave the  update_attribute alone since that can be used internally.

app/models/miq_ae_domain.rb
```diff
@@ -100,6 +100,17 @@ def editable_property?(property)
 
   alias editable_contents? editable?
 
+  def update_attributes_api(hash)
```
gmcculloug: @lfu @mkanoor Do we know where the API is calling this?  If it is generic logic on the API side it might be difficult to introduce a different method name.

lfu: @skateman Any idea?

skateman: Yeah, that method name is probably not okay, cc @martinpovolny @abellotti

lfu: @skateman Do you know where API is calling this? Can you post the link here?

skateman: I don't think that the API would be able call this, but if the method is named correctly, we can work with it. Others might know more ...

lfu: @skateman Sorry did not make it clear. Regardless of the method name, where would the new method be called in API?

lfu: Can we name it `update_properties`?

tinaafitz:
@martinpovolny @skateman, fyi - @mkanoor, @lfu and I are meeting with @Fryguy this afternoon to discuss these changes.

gmcculloug:
@tinaafitz @lfu Do we still need this PR or can it be closed?

skateman:
The question is more to @Fryguy I guess, do we want an extra layer of security on top?

lfu:
This can be closed because we will do it in a different way.

Issue #385, opened 2019-10-30
Modify YAML for Service Model encoding/decoding to store minimal object properties


gmcculloug:
Modify the data that YAML encodes for a service model object to just the class name and object ID, which is the ID of the backing ActiveRecord object.

Example:
https://github.com/ManageIQ/manageiq/blob/0d607394a7a7334d1923f1e711037b5b694383d7/lib/extensions/ar_yaml.rb#L6-L15

Also see: https://stackoverflow.com/questions/10629209/how-can-i-control-which-fields-to-serialize-with-yaml

gmcculloug:
@pkomanek Talking to @mkanoor he was suggesting we use something like the serialize/deserialize technique used here (https://github.com/ManageIQ/manageiq-automation_engine/pull/64) to convert the state_vars.

On `set_state_var` we would also want to run the serialization and raise an error if an unsupported object class was passed.  We need to work on exactly what the supported class list would but, but assume the classes supported by [YAML.safe_load](https://ruby-doc.org/stdlib-2.6.1/libdoc/psych/rdoc/Psych.html#method-c-safe_load) plus a few others, like Symbol, Date and the Service Models.  (Service models would be referenced with a simple string format)

Take a look and we can discuss more.

gmcculloug:
Resolved by https://github.com/ManageIQ/manageiq-automation_engine/pull/387 and https://github.com/ManageIQ/manageiq-automation_engine/pull/398

PR #387, opened 2019-11-06
Reducing the amount of data being stored in yaml file for svc_models

pkomanek:
This PR modifies the amount of data for automate service models YAML export. From now we will store only the class name and ID of related active_record object. The rest of the information will be loaded from related ar_model.

Based on https://github.com/ManageIQ/manageiq-automation_engine/issues/385

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -367,6 +367,15 @@ module MiqAeMethodService
       self.class == other.class && id == other.id
     end
 
+    def encode_with(coder)
+      coder['id'] = id
+    end
+
+    def init_with(coder)
+      @object = self.class.service_model_name_to_model(self.class.name)&.find_by(:id => coder['id'])
+      self
+    end
+
     private
 
     def verify_taggable_model
--- a/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
@@ -94,6 +94,30 @@ describe MiqAeMethodService::MiqAeServiceModelBase do
     end
   end
 
+  describe 'YAML import/export' do
+    let(:service)     { FactoryBot.create(:service, :name => 'test_service') }
+    let(:svc_service) { MiqAeMethodService::MiqAeServiceService.find(service.id) }
+
+    it 'exports only class name and ID' do
+      expect(svc_service.to_yaml)
+        .to eq("--- !ruby/object:#{svc_service.class.name}\nid: #{svc_service.id}\n")
+    end
+
+    it 'loads object from yaml' do
+      expect(YAML.safe_load(svc_service.to_yaml)).to eq(svc_service)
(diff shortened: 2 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -367,6 +367,16 @@ def ==(other)
       self.class == other.class && id == other.id
     end
 
+    def encode_with(coder)
+      coder['id'] = id
+    end
+
+    def init_with(coder)
+      klass = self.class.to_s.split("MiqAeMethodService::MiqAeService")[1].constantize
+      instance_variable_set('@object', klass.find(coder['id']))
```
gmcculloug: This can be reduced to:
```ruby
@object = service_model_name_to_model(self.class.name).find_by(:id => coder['id'])
```

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -94,6 +94,28 @@
...
+    it 'loads object from yaml' do
+      expect(YAML.safe_load(svc_service.to_yaml)).to eq(svc_service)
+    end
+
+    it 'returns nil for objects without related active record instance' do
+      yaml = svc_service.to_yaml
+      service.delete
+      # This test and tested behaviour is in-progress state. Right now we are
+      # getting back the invalid service_model instance with blank @object variable.
+      expect(YAML.safe_load(yaml)).to eq(nil)
```
gmcculloug: @pkomanek I do not think we will be able to accomplish this which means this approach is not going to be the right solution.  In discussing this issue with @mkanoor and @tinaafitz we talked about going in a different direction to solve the original problem.

Suggesting we close this PR for now and I will update the git issue with more details.

tinaafitz:
@gmcculloug @mkanoor Please review.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -367,6 +367,19 @@ def ==(other)
...
+      coder['id'] = id
+    end
+
+    def init_with(coder)
+      @object = self.class.service_model_name_to_model(self.class.name)&.find_by(:id => coder['id'])
+      self
+    end
+
+    def ar_model_present?
+      !@object.nil?
```
gmcculloug: @pkomanek Use `@object.present?`

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -94,6 +94,21 @@
...
+    it 'loads invalid svc_model for objects without related ar_model' do
+      yaml = svc_service.to_yaml
+      service.delete
+      model_from_yaml = YAML.safe_load(yaml)
+      expect { model_from_yaml.reload }.to raise_error(
+        NoMethodError,
+        "undefined method `reload' for nil:NilClass"
+      )
+    end
+  end
```
gmcculloug: Looks like we lost some other tests that were still meaningful for testing the encode/init changes.

PR #390, opened 2019-11-16
[HAMMER] Fix manageiq-content specs on hammer

NickLaMuro:
(Note:  Marked as `[WIP]` until it is determined that https://github.com/ManageIQ/manageiq-content/pull/559 is safe to stay in `hammer` as is [ [ref](https://github.com/ManageIQ/manageiq-content/pull/559#issuecomment-554580025) ] )

The following PR was recently backported to `hammer`:

https://github.com/ManageIQ/manageiq-content/pull/559

This made use of the `config_info` association to accomplish its goals.  It was possible to use this method without the changes made here since a PR was merged and available as of `ivanchuk` that exposed all service associations to Miq AE service models:

https://github.com/ManageIQ/manageiq-automation_engine/pull/308

But that PR was not merged to `hammer`.  Do to the risk of backporting that PR, this is a one off addition to allow the functionality of ManageIQ/manageiq-content#559 (low risk) to function without requiring a backport of everything in ManageIQ/manageiq-automation_engine#308 (high risk).

Links
-----

* Failing test run:  https://travis-ci.org/ManageIQ/manageiq-content/builds/612036937
* Offending PR: https://github.com/ManageIQ/manageiq-content/pull/559
* PR allowing the above to work on `master`/`ivanchuk`:  https://github.com/ManageIQ/manageiq-automation_engine/pull/308

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_service_template_ansible_playbook.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_service_template_ansible_playbook.rb
@@ -1,5 +1,6 @@
 module MiqAeMethodService
   class MiqAeServiceServiceTemplateAnsiblePlaybook < MiqAeServiceServiceTemplateGeneric
+    expose :config_info, :association => true
     expose :job_template
   end
 end


NickLaMuro:
@miq-bot add_label test, bug, hammer/yes

NickLaMuro:
@d-m-u There... _**it's**_ fixed...

d-m-u:
Lucy says this only fixes this specific case though, I think?

NickLaMuro:
> Lucy says this only fixes this specific case though, I think?

@d-m-u With a build coming up for `hammer`, and this being a backport, that is exactly how we should be addressing this, and fixing it to allow the "specific case" to work.

It is "low risk" in that case, and doing the higher risk thing should be left for major releases.

lfu:
I agree with what @NickLaMuro said. This PR looks good to me. 
`virtual_has_one` seems never worked with Hammer. It would take time to fix and test it.

simaishi:
So... are we good un-wip'ing this PR? I think @lfu confirmed https://github.com/ManageIQ/manageiq-content/pull/559 can stay in hammer branch.

tinaafitz:
@simaishi, I'm in favor of un-wip'ing it.

NickLaMuro:
> I agree with what @NickLaMuro said. This PR looks good to me. virtual_has_one seems never worked with Hammer

@lfu I want to correct this line of thinking:  `virtual_has_one` was never broken.  The reason that `ivanchuk`/`master` work without this change is because of https://github.com/ManageIQ/manageiq-automation_engine/pull/308 being include in those two branches.

That code specifically added this is here:

https://github.com/ManageIQ/manageiq-automation_engine/blob/e67731df2db2e2d442c126b12e7f3ae6c7d27ed2/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb#L45-L56

(and part of the diff [here](https://github.com/ManageIQ/manageiq-automation_engine/pull/308/files#diff-81cc1c401f00f38283f3f78b9373bc87) )

This basically did the code in this PR ( `expose(key, :association => true)` ) automatically, which  since it is a `virtual_has_one`, not a `attribute`, it always required the extra `:association => true` bit.

There is nothing "broken" with `virtual_has_one` in this respect, just expectations about what is automatically `expose`d by default in `miq_ae_service_model_base.rb`.  In `hammer`, only public `attributes` are exposed automatically by default.

d-m-u:
Thanks for the clarification @NickLaMuro

simaishi:
Thank you everyone!

lfu:
> In `hammer`, only public `attributes` are exposed automatically by default.

I should have made my statement more specific: in `hammer`, `virtual_has_one` has never automatically exposed the associations. But both public attributes and `virtual_column` are exposed automatically.

PR #394, opened 2019-12-04
Disable streaming_refresh in MiqAeServiceEmsEvent spec and fix CI

lpichler:
In those tests, we are testing whether method  `EmsRefresh.queue_refresh` was called with certain parameters when method  [`EmsEvent::Automate.refresh`](https://github.com/ManageIQ/manageiq-automation_engine/blob/baeea09c1241fefda279ebeff139177040ccbca5/spec/service_models/miq_ae_service_ems_event_spec.rb#L24) is called.

`EmsEvent::Automate.refresh` has [guard condition](https://github.com/ManageIQ/manageiq/blob/c79c6219d51dbb8aaa385a2acd3c28adfda3a739/app/models/ems_event/automate.rb#L14) and when `Settings.ems_refresh.vmwarews.streaming_refresh = true` - no code from `EmsEvent::Automate.refresh`  is executed and there is method `EmsRefresh.queue_refresh` called and we call the method for test.

Before https://github.com/ManageIQ/manageiq-providers-vmware/pull/434 we were using `Settings.prototype.ems_vmware.update_driven_refresh` and it was to `nil` or `false` and then guard did not apply.

## Links
CI failure: https://travis-ci.com/ManageIQ/manageiq-cross_repo-tests/jobs/263025345 
Caused by: https://github.com/ManageIQ/manageiq-providers-vmware/pull/434

@miq-bot add_label test

@miq-bot assign @gmcculloug

### DIFF
--- a/spec/service_models/miq_ae_service_ems_event_spec.rb
+++ b/spec/service_models/miq_ae_service_ems_event_spec.rb
@@ -1,5 +1,7 @@
 describe MiqAeMethodService::MiqAeServiceEmsEvent do
   before do
+    Settings.ems_refresh.vmwarews.streaming_refresh = false
+
     @ems           = FactoryBot.create(:ems_vmware_with_authentication,
                                         :zone => FactoryBot.create(:zone)
                                        )


spec/service_models/miq_ae_service_ems_event_spec.rb
```diff
@@ -1,5 +1,7 @@
 describe MiqAeMethodService::MiqAeServiceEmsEvent do
   before do
+    Settings.ems_refresh.vmwarews.streaming_refresh = false
```
gmcculloug: Is this going to cause cross-test corruption.  Does this value need to be reset to the initial value in an `after` block?

agrare: `stub_settings_merge(:ems_refresh => {:vmwarews => {:streaming_refresh => false}})` would be the better way to do this, but it won't cause test contamination afaik

agrare:
@gmcculloug streaming refresh will be the only method soon (https://github.com/ManageIQ/manageiq-providers-vmware/pull/488)

I'll make sure to update this test before that gets merged

gmcculloug:
Closing in favor of https://github.com/ManageIQ/manageiq-automation_engine/pull/396

PR #396, opened 2019-12-06
Test an EMS that doesn't use streaming refresh

agrare:
Test the EmsEvent refresh handler with an EMS that doesn't use streaming
refresh.

Alternative to: https://github.com/ManageIQ/manageiq-automation_engine/pull/394


agrare:
@gmcculloug eventually there won't be an option for streaming/legacy refresh for VMware, so this switches the type of EMS being tested

PR #398, opened 2019-12-12
Warning message for imported service models related to non-existing ar_models

pkomanek:
Adding the warning message for importing service models for non-existing active record objects. PR also includes the name changing for one of the service model method from 'ar_model_present?' to '#record_exists?'

this PR is based on https://github.com/ManageIQ/manageiq-automation_engine/issues/385#issuecomment-553672232

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -360,6 +360,8 @@ module MiqAeMethodService
     end
 
     def ar_method(&block)
+      return if @object.nil?
+
       self.class.ar_method(&block)
     end
 
@@ -373,6 +375,8 @@ module MiqAeMethodService
 
     def init_with(coder)
       @object = self.class.service_model_name_to_model(self.class.name)&.find_by(:id => coder['id'])
+      $miq_ae_logger.warn("There is no related active record object with id=#{coder['id']} for imported #{self.class}") if @object.nil?
+
       self
     end
 
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -320,6 +320,8 @@ module MiqAeMethodService
     end
 
     def reload
+      raise ActiveRecord::RecordNotFound, "Couldn't find related ActiveRecord object" unless record_exists?
+
       object_send(:reload)
       self # Return self to prevent the internal object from being returned
     end
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -380,6 +380,10 @@ module MiqAeMethodService
       self
     end
 
+    def record_exists?
+      @object.present?
+    end
+
     private
 
     def verify_taggable_model
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
@@ -263,7 +263,7 @@ module MiqAeMethodService
         raise ArgumentError, "#{ar_klass.name} Object expected, but received #{obj.class.name}"
       end
 
-      @object = obj.kind_of?(ar_klass) ? obj : ar_method { ar_klass.find_by(:id => obj) }
+      @object = obj.kind_of?(ar_klass) ? obj : self.class.ar_method { ar_klass.find_by(:id => obj) }
       raise MiqAeException::ServiceNotFound, "#{ar_klass.name} Object [#{obj}] not found" if @object.nil?
     end
 
--- a/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
@@ -111,10 +111,8 @@ describe MiqAeMethodService::MiqAeServiceModelBase do
       yaml = svc_service.to_yaml
       service.delete
       model_from_yaml = YAML.safe_load(yaml, [MiqAeMethodService::MiqAeServiceService])
-      expect { model_from_yaml.reload }.to raise_error(
-        NoMethodError,
-        "undefined method `reload' for nil:NilClass"
-      )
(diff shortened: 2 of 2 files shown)


pkomanek:
@gmcculloug the '#record_exists?' method is working fine for the imported service_models. But if you do this scenario: 
```
service = Service.create(....)
MiqAeMethodService::MiqAeServiceService.find(....)
service.delete
svc_service.record_exists?
```
the result will be true, because the instance variable `@object` is present.  I am not sure, if this is the correct behaviour. Maybe we should check the presents of the `@object` and the do the reload to be sure, that the related `ar_model` exists. What do you think about it ?

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -373,9 +373,15 @@ def encode_with(coder)
 
     def init_with(coder)
       @object = self.class.service_model_name_to_model(self.class.name)&.find_by(:id => coder['id'])
+      $miq_ae_logger.warn("There is no related active record object with id='#{coder[id]}' for imported #{self.class}") if @object.nil?
```
gmcculloug: We do not need the quotes around the `coder['id'])` value when logging and note that the log message is passing `coder[id]` instead of `coder['id']`

```ruby
$miq_ae_logger.warn("There is no related active record object with id=#{coder['id']} for imported #{self.class}") if @object.nil?
```

gmcculloug:
@pkomanek I was reviewing the discussion we were having in chat around these changes and the error raised when the active_record object no longer exists.  Mainly this:

> I am also sending you piece of the log after importing the svc_model with relation to non-existing ar_model.
```
[----] W, [2019-12-19T17:17:04.767496 #11179:2ae1b49365f0]  WARN -- : There is no related active record object with id='3658' for loaded 'MiqAeMethodService::MiqAeServiceMiqAeDomain'
[----] E, [2019-12-19T17:17:07.312126 #11179:2ae1b49365f0] ERROR -- : MiqAeServiceModelBase.ar_method raised: <NoMethodError>: <undefined method `virtual_attribute_names' for NilClass:Class>
[----] E, [2019-12-19T17:17:07.312255 #11179:2ae1b49365f0] ERROR -- : /home/pkomanek/manageiq/manageiq-automation_engine/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb:2
```

I think the solution here would be to add the following to `ar_method` in the `miq_ae_service_model_base.rb` file you are already modifying.  This protects the code from calling methods that we know will not succeed.

```diff
     def ar_method(&block)
+      return if @object.nil?
+
       self.class.ar_method(&block)
     end
```

gmcculloug:
@pkomanek Any thoughts on the recommended approach above?

gmcculloug:
@pkomanek Please review the failing tests.  I understand the one `reload` error but most are failing on `ServiceNotFound`.

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -111,10 +111,8 @@
       yaml = svc_service.to_yaml
       service.delete
       model_from_yaml = YAML.safe_load(yaml, [MiqAeMethodService::MiqAeServiceService])
-      expect { model_from_yaml.reload }.to raise_error(
-        NoMethodError,
-        "undefined method `reload' for nil:NilClass"
-      )
```
gmcculloug: @pkomanek As discussed we still want an error from the `reload` method when the `record_exists?` is false.  Similar to what we see when active_record reloads an object with YAML but the record does not exist in the db.

Here is the ruby script I used to prove this out (and the great writeup that explains why this happens [active_weirdness](https://gist.github.com/NickLaMuro/11b2119819dca4408c5510831c368be5)):
```ruby
# Create new AR object
s = Service.create!(:name => "delete me")
s.reload
s.id
# => 389

# Save it to disk as YAML and destroy the object
File.write('test.yaml', s.to_yaml)
s.destroy

# Load the object from the saved file (you can restart console here to ensure it is not in memory)
s2 = YAML.safe_load(File.read('test.yaml'))
# Save ID as before
s2.id
# => 389

# Reload shows the object is not in the db
s2.reload
# => ActiveRecord::RecordNotFound (Couldn't find Service with 'id'=389)

# But save! and update! seem to work even though the object only existin in memory
s2.save!
#  (0.4ms)  BEGIN
# MiqGroup Load (0.7ms)  SELECT  "miq_groups".* FROM "miq_groups" WHERE "miq_groups"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
# MiqGroup Inst Including Associations (18.8ms - 1rows)
# Tenant Load (1.2ms)  SELECT  "tenants".* FROM "tenants" WHERE "tenants"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
# Tenant Inst Including Associations (19.6ms - 1rows)
# Service Exists (0.6ms)  SELECT  1 AS one FROM "services" WHERE "services"."guid" = $1 AND ("services"."id" != $2) LIMIT $3  [["guid", "308ee870-17cc-49ee-b288-64b2c22b63e1"], ["id", 389], ["LIMIT", 1]]
#  (0.3ms)  COMMIT
# => true

s2.update!(:name => "please delete me")
#  (0.3ms)  BEGIN
# Service Exists (0.7ms)  SELECT  1 AS one FROM "services" WHERE "services"."guid" = $1 AND ("services"."id" != $2) LIMIT $3  [["guid", "308ee870-17cc-49ee-b288-64b2c22b63e1"], ["id", 389], ["LIMIT", 1]]
# SQL (0.5ms)  UPDATE "services" SET "name" = $1, "updated_at" = $2 WHERE "services"."id" = $3  [["name", "please delete me"], ["updated_at", "2019-10-22 21:29:13.009851"], ["id", 389]]
#  (0.2ms)  COMMIT
# => true

# Reload again shows nothing is actually persisted
s2.reload
# ActiveRecord::RecordNotFound (Couldn't find Service with 'id'=389)
```

pkomanek: @gmcculloug I modified your scenario for our case:

```
s = Service.create!(:name => "delete me")
s.reload
svc_s = MiqAeMethodService::MiqAeServiceService.find(s.id)
yaml = svc_s.to_yaml
s.destroy
svc_s2 = YAML.safe_load(yaml)
svc_s2.reload
# => ActiveRecord::RecordNotFound (Couldn't find related ActiveRecord object)
svc_s.reaload
# => ActiveRecord::RecordNotFound (Couldn't find Service with 'id'=...)
```

gmcculloug:
@mkanoor Please review

PR #405, opened 2019-12-31
State var yaml to binary blob

gmcculloug:
This PR creates a new class `StateVarHash` which extends the `HashWithIndifferentAccess` normally used to store the state_var hash data.  

The key impact of this change is that the new class overrides the YAML dump/load logic and will create a binary_blob record used to store the hash data.  When automate performs a retry the state_var hash will add minimal YAML data to the MiqQueue record instead of serializing the YAML representation of the state_var hash.  (If the state_var is an empty hash, no binary blob record will be created and the `binary_blob_id` will be set to 0.)

Reducing the data stored on the queue to something like:
```yaml
--- !ruby/object:MiqAeEngine::StateVarHash
binary_blob_id: <####>
```

Followup PR https://github.com/ManageIQ/manageiq-automation_engine/pull/406 limits the state_var key to a few simple scalar types.

The first commit is the core changes to add the class and implement it, the second commit is only spec changes.

Existing spec tests cover the serialization of the state_vars during an automate retry.
[spec/engine/miq_ae_state_machine_retry_spec.rb#L273](https://github.com/ManageIQ/manageiq-automation_engine/blob/03cbde7b3cc682f2a32b92009fa736c17e97f848/spec/engine/miq_ae_state_machine_retry_spec.rb#L273)
[spec/engine/miq_ae_state_machine_retry_spec.rb#L185](https://github.com/ManageIQ/manageiq-automation_engine/blob/03cbde7b3cc682f2a32b92009fa736c17e97f848/spec/engine/miq_ae_state_machine_retry_spec.rb#L185)

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/state_var_hash.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/state_var_hash.rb
@@ -1,0 +1,27 @@
+module MiqAeEngine
+  class StateVarHash < HashWithIndifferentAccess
+    SERIALIZE_KEY = 'binary_blob_id'.freeze
+
+    def encode_with(coder)
+      coder[SERIALIZE_KEY] =
+        blank? ? 0 : BinaryBlob.new.tap { |bb| bb.store_data("YAML", to_h) }.id
+    end
+
+    def init_with(coder)
+      return if coder[SERIALIZE_KEY].zero?
+
+      binary_blob = BinaryBlob.find_by(:id => coder[SERIALIZE_KEY])
+      if binary_blob
+        blob_hash = binary_blob.data
+        binary_blob.destroy
+
+        $miq_ae_logger.info("Reloading state var data: #{blob_hash.to_yaml}")
+        update(blob_hash)
+      else
+        $miq_ae_logger.warn("Failed to load BinaryBlob with ID [#{coder[SERIALIZE_KEY]}] for #{self.class.name}")
+      end
+
+      self
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/state_var_hash.rb
```diff
@@ -0,0 +1,24 @@
+module MiqAeEngine
+  class StateVarHash < HashWithIndifferentAccess
+    SERIALIZE_KEY = 'binary_blob_id'.freeze
+
+    def encode_with(coder)
+      coder[SERIALIZE_KEY] =
+        blank? ? 0 : BinaryBlob.new.tap { |bb| bb.store_data("YAML", to_h) }.id
```
mkanoor: @gmcculloug should this be segregated based on tenant?

gmcculloug:
cc @Fryguy

lib/miq_automation_engine/engine/miq_ae_engine/state_var_hash.rb
gmcculloug: Considering the BinaryBlob model does not have a tenant column and this value is never exposed external I do not see that being a concern.

mkanoor: @gmcculloug One thing we are going to be missing out after this change is the ability to examine what data the user has been using for state vars. Previously these would get logged in the log would it make sense to log the hash as soon as we decode it so we know what the data looks like.

gmcculloug: Interesting, I had not considered the logging side of things, but that should be easy to add.  I'll try to get that in later today.

gmcculloug: @mkanoor I pushed changes to log the hash (as a yaml string) and added tests to validate that we log on both successful and failed conditions in the `init_with` method.

gmcculloug:
@mkanoor PR updated with logging.  PTAL

PR #406, opened 2020-01-04
State var limit key type

gmcculloug:
Dependent on https://github.com/ManageIQ/manageiq-automation_engine/pull/405

This PR extends the `StateVarHash` class to limit what class types can be used as state var keys.

The current list includes: `NilClass, Numeric, String, Symbol, Date, Time`.   (Open to other suggestion, but the idea is to limit it to basic object types.)


gmcculloug:
cc @pemcg

gmcculloug:
@pemcg Would love to get your thoughts on this change.  Is the current list of classes (`NilClass, Numeric, String, Symbol, Date, Time`) sufficient or are the other classes you would suggest adding?

pemcg:
Looks good to me, I cannot think of any other key types that would be useful

gmcculloug:
@tinaafitz Remove the `WIP` flag and should be ready to merge now.

PR #407, opened 2020-01-09
Use YAML.load to load classes beyond the basic types

jrafanie:
https://bugzilla.redhat.com/show_bug.cgi?id=1789153

Followup to ManageIQ/manageiq#19701

The prior behavior in core was to treat YAML.safe_load like YAML.load so let's change
some of these to .load for now.  We'll enumerate the list of classes where we
can.

Also, fix typo of backtick instead of quote.

### DIFF
--- a/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
+++ b/spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
@@ -104,13 +104,13 @@ describe MiqAeMethodService::MiqAeServiceModelBase do
     end
 
     it 'loads object from yaml' do
-      expect(YAML.safe_load(svc_service.to_yaml)).to eq(svc_service)
+      expect(YAML.safe_load(svc_service.to_yaml, [MiqAeMethodService::MiqAeServiceService])).to eq(svc_service)
     end
 
     it 'loads invalid svc_model for objects without related ar_model' do
       yaml = svc_service.to_yaml
       service.delete
-      model_from_yaml = YAML.safe_load(yaml)
+      model_from_yaml = YAML.safe_load(yaml, [MiqAeMethodService::MiqAeServiceService])
       expect { model_from_yaml.reload }.to raise_error(
         NoMethodError,
         "undefined method `reload' for nil:NilClass"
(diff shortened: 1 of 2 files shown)


spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -104,13 +104,13 @@
     end
 
     it 'loads object from yaml' do
-      expect(YAML.safe_load(svc_service.to_yaml)).to eq(svc_service)
+      expect(YAML.load(svc_service.to_yaml)).to eq(svc_service)
```
gmcculloug: Suggest we pass `MiqAeMethodService::MiqAeServiceService` to safe_load on this one.

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -104,13 +104,13 @@
...
     end
 
     it 'loads invalid svc_model for objects without related ar_model' do
       yaml = svc_service.to_yaml
       service.delete
-      model_from_yaml = YAML.safe_load(yaml)
+      model_from_yaml = YAML.load(yaml)
```
gmcculloug: Same here,  Pass `MiqAeMethodService::MiqAeServiceService` to safe_load.

spec/engine/miq_ae_state_machine_retry_spec.rb
```diff
@@ -280,6 +280,6 @@ def create_restart_model(script1, script2, script3)
     expect(MiqQueue.count).to eq(2)
     q = MiqQueue.where(:state => 'ready').first
     expect(q[:server_guid]).to be_nil
-    expect(YAML.safe_load(q.args.first[:ae_state_data])).to eq(ae_state_data)
+    expect(YAML.load(q.args.first[:ae_state_data])).to eq(ae_state_data)
```
gmcculloug: This one is better off using `load` for now.

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
jrafanie: I just wanted to see you say Service Service Service 🤣

jrafanie: We might need HashWithIndifferentAccess too.  I'll check.

jrafanie: Nope, just Service ServiceService was enough

PR #416, opened 2020-02-04
Fix set_vm_notes spec

d-m-u:
~~Can we nix this? It's not like it's doing anything.~~ 
This was commented out, as GM says, it could be a valid test.

Something like this?

### DIFF
--- a/spec/service_models/miq_ae_service_miq_provision_spec.rb
+++ b/spec/service_models/miq_ae_service_miq_provision_spec.rb
@@ -110,13 +110,12 @@ describe MiqAeMethodService::MiqAeServiceMiqProvision do
   end
 
   it "#set_vm_notes" do
-    method   = "$evm.root['#{@ae_result_key}'] = $evm.root['miq_provision'].set_vm_notes"
+    method = "$evm.root['#{@ae_result_key}'] = $evm.root['miq_provision'].set_vm_notes('homer')"
     @ae_method.update(:data => method)
-
-    # %w{ template clone_to_vm clone_to_template }.each do |provision_type|
-    #  @miq_provision.update(:provision_type => provision_type)
-    #  invoke_ae.root(@ae_result_key).should == @miq_provision.provision_type
-    # end
+    expect(invoke_ae.root(@ae_result_key)).to be_truthy
+    expect(@miq_provision[:options][:vm_notes]).to be_nil
+    @miq_provision.reload
+    expect(@miq_provision[:options][:vm_notes]).to eq('homer')
   end
 
   it "#target_type" do


spec/service_models/miq_ae_service_miq_provision_spec.rb
```diff
@@ -109,16 +109,6 @@ def invoke_ae
     expect(callback_hash[:next_time_around]).to eq('do_something_better_yet')
   end
 
-  it "#set_vm_notes" do
-    method   = "$evm.root['#{@ae_result_key}'] = $evm.root['miq_provision'].set_vm_notes"
-    @ae_method.update(:data => method)
```
gmcculloug: What if we instead keep the test and changed it to call `set_vm_notes` on the service model and ensure that the proper backend method is called.

d-m-u:
It's the same failure as on the just-merged changelog PR.

gmcculloug:
@d-m-u Changelog PRs typically use `[skip ci]` since they are documentation changes only so the PR you referenced was green when it was merged.  We should not be ignoring failing tests.

I'll get some time this morning to review the test failures.  Also letting @tinaafitz and @lfu know about it.

spec/service_models/miq_ae_service_miq_provision_spec.rb
```diff
@@ -110,13 +110,12 @@ def invoke_ae
   end
 
   it "#set_vm_notes" do
-    method   = "$evm.root['#{@ae_result_key}'] = $evm.root['miq_provision'].set_vm_notes"
+    method = "$evm.root['#{@ae_result_key}'] = $evm.root['miq_provision'].set_vm_notes('bite_me')"
```
gmcculloug: I'll assume you did this just to see if I was paying attention.  Let's try again.   If you need a suggestion for a string to pass please let me know.

spec/service_models/miq_ae_service_miq_provision_spec.rb
```diff
@@ -110,13 +110,12 @@ def invoke_ae
...
     @ae_method.update(:data => method)
-
-    # %w{ template clone_to_vm clone_to_template }.each do |provision_type|
-    #  @miq_provision.update(:provision_type => provision_type)
-    #  invoke_ae.root(@ae_result_key).should == @miq_provision.provision_type
-    # end
+    expect(invoke_ae.root(@ae_result_key)).to be_truthy
+    expect(@miq_provision[:options][:vm_notes]).to be_nil
+    @miq_provision.reload
+    expect(@miq_provision[:options][:vm_notes]).to eq('bite_me')
```
gmcculloug: I prefer `@miq_provision.options` over the `[]` reference, but this seems to be consistent with other lines in the PR.  Keep it this way for the PR and we can clean up in a followup refactoring.

spec/service_models/miq_ae_service_miq_provision_spec.rb
gmcculloug: Nice Iliad reference, it's normally uppercased, but I'll take it.  😉

PR #423, opened 2020-02-07
Change options accessing syntax in spec 

d-m-u:
`@miq_provision[:options]` should use `@miq_provision.options` syntax

per https://github.com/ManageIQ/manageiq-automation_engine/pull/416#discussion_r376386419


gmcculloug:
@d-m-u Thanks

PR #429, opened 2020-03-04
Remove the to_xml call from miq_ae_engine

d-m-u:
Per needinfo on https://bugzilla.redhat.com/show_bug.cgi?id=1805451 I was asked to remove the call to_xml as it's old and unused now. 

@miq-bot assign @tinaafitz 

We I guess have a plan to remove the rest of them sometime soon but for now this should be sufficient for the bz.

Issue #418, opened 2020-02-05
Investigate replacing `require_relative` with `require_dependency`


gmcculloug:
Based on work from https://github.com/ManageIQ/manageiq-automation_engine/pull/415 `Guard against hitting const_missing from within const_missing` 

See https://github.com/ManageIQ/manageiq-automation_engine/pull/415#discussion_r374951896

skateman:
**Before code reload:**
![Screenshot from 2020-03-13 12-44-04](https://user-images.githubusercontent.com/649130/76618347-afc14e80-6528-11ea-9733-9ebe3eb08127.png)

**After code reload:**
![Screenshot from 2020-03-13 12-45-29](https://user-images.githubusercontent.com/649130/76618355-b3ed6c00-6528-11ea-9fe3-dd78435e6052.png)

This is just a demonstration how the automation results are altered after a code reload. It's nearly impossible to do any work in the area :confused:

jrafanie:
I'm looking at this now as the issue in https://github.com/ManageIQ/manageiq-automation_engine/pull/415#discussion_r374744630 indicates the reloading of code in dev mode isn't properly reloading `MiqAeServiceModelBase`

jrafanie:
@pkomanek I'm going to do some testing here and let you know what I find. So far, it looks like either the #415 change was incomplete OR that when rails reloads code, it fails to load it correctly, or both.  

Either way, after reload `MiqAeServiceModelBase` doesn't contain `create_service_model_from_name` method, so it blows up:

```
[----] I, [2020-03-13T18:09:14.247656 #83183:3fc8c875e3f8]  INFO -- : XXX: convert_value_based_on_datatype 1 "MiqServer"
[----] I, [2020-03-13T18:09:14.252759 #83183:3fc8c875e3f8]  INFO -- : XXX const_missing 1 name: :MiqAeServiceMiqServer
[----] I, [2020-03-13T18:09:14.252843 #83183:3fc8c875e3f8]  INFO -- : XXX const_missing 2 defined?(MiqAeServiceModelBase): nil
[----] I, [2020-03-13T18:09:14.252879 #83183:3fc8c875e3f8]  INFO -- : XXX const_missing 3
[----] I, [2020-03-13T18:09:14.252911 #83183:3fc8c875e3f8]  INFO -- : XXX const_missing 1 name: :MiqAeServiceModelBase
[...]

PR #430, opened 2020-03-05
[WIP] Remove to_xml

d-m-u:
This is follow-up to https://github.com/ManageIQ/manageiq-automation_engine/pull/429.

This was what I could remove without specs failing, and it's kinda late, so I'll get back to it tomorrow.

there are a few more:
``` ruby
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:193:    def attribute_value_to_xml(value, xml)
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:204:            xml.Element(:index => i + 1) { attribute_value_to_xml(value[i], xml) }
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:210:            xml.Key(:name => k.to_s) { attribute_value_to_xml(v, xml) }
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:221:    def to_xml(options = {})
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:227:          xml.MiqAeAttribute(:name => k) { attribute_value_to_xml(@attributes[k], xml) }
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb:230:        children.each { |c| c.to_xml(:builder => xml) }
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb:201:        objs.each { |obj| obj.to_xml(:builder => xml) }
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb:205:    def to_xml(path = nil)
lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb:209:      XmlHash.from_hash({"MiqAeObject" => result}, {:rootname => "MiqAeWorkspace"}).to_xml.write(s, 2)
```

but with those removed as well, https://github.com/ManageIQ/manageiq-automation_engine/blob/1a3ba3a3cde0706890050e405f56b4c808213c1e/spec/miq_ae_state_machine_spec.rb#L37 fails with `ae_result` of `error` rather than `ok`.

Maybe also `app/models/miq_ae_datastore/xml_import.rb` can go, but I have to look closer at it tomorrow cause it causes ` uninitialized constant` constant problems


d-m-u:
@miq-bot add_label technical debt

gmcculloug:
@mkanoor Any concern with trying to get rid of all the XML stuff on master?  It likely needs to happen over the course of a few PRs, but feels like a good technical debt cleanup.  Thoughts?

mkanoor:
@gmcculloug I think the UI uses this to display the output of simulation, we would have to coordinate with them to start using JSON instead of XML when displaying the simulation output.
https://github.com/ManageIQ/manageiq-ui-classic/blob/b0e5666b3d3abfa857856c4bae70759039f84c6c/app/presenters/tree_builder_automate_simulation_results.rb#L17
There might be other places in the UI. Otherwise looks good

skateman:
The Automate->Simulation in the UI depends on the `to_xml` method as the simulation results can be displayed as XML and the tree view is being built by parsing the same XML. I can try to build the tree directly from the result object, but it doesn't really help if we don't get rid of the XML view.

Screenshot-from-2020-03-11-11-15-10

gmcculloug:
@skateman I see little use for the `XML View` tab so I feel it is safe to remove.  @mkanoor Do we have the methods to support building the `Tree View` from YAML or JSON?

skateman:
@gmcculloug we don't really need the YAML or JSON version, I can just build the tree directly from the objects. Faster to build directly and avoid parsing of formats...

d-m-u:
there's also `to_expanded_xml` which I'm not sure if I can remove or not because we use it in ```miq_log_workspace(obj, _inputs)``` inside miq_ae_builtin_method

skateman:
That's the one I'm dependent on in the UI, but it's getting close...

d-m-u:
gotcha, thanks

skateman:
I'm blocked by https://github.com/ManageIQ/manageiq-automation_engine/issues/418 but @jrafanie is working on it

d-m-u:
it's fine, this's got that failing spec that i'm looking into, there's no rush

skateman:
Orienting in the code is way too complicated for me, so maybe @mkanoor's idea about you implementing a `to_json` version isn't that bad...if you can.

lfu:
How about [xml_import](https://github.com/ManageIQ/manageiq-automation_engine/blob/ee70c81750d70c7e25a992b638b619b09868b70b/app/models/miq_ae_datastore/xml_import.rb#L2)? Do we not remove it this time?

djberg96:
@skateman, @lfu is the baked in Rails method something we could use?

https://apidock.com/rails/Hash/to_xml

PR #201, opened 2018-07-23
Start the drb server with a unix socket

jrafanie:
Use a temp file location and set permissions so only the current system
user can access it.

CVE-2018-10905
https://bugzilla.redhat.com/show_bug.cgi?id=1599389

PR #431, opened 2020-03-13
Explicitly pass a local URI for our client's DRb.start_service

jrafanie:
While testing some workspace instantiantion at home, I was seeing this in the
logs:

```
[----] E, [2020-03-12T16:35:06.554957 #39577:3ff30d01cf44] ERROR -- : <AEMethod available_resource_groups>   DRb::DRbConnError: druby://92.242.140.21:54933 - #<Errno::ETIMEDOUT: Operation timed out - connect(2) for "92.242.140.21" port 54933>
[----] E, [2020-03-12T16:35:06.556663 #39577:3ff30d01cf44] ERROR -- : <AEMethod available_resource_groups>   (drbunix:///var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine20200312-39577-6nm7ah) /Users/joerafaniello/.rubies/ruby-2.6.5/lib/ruby/2.6.0/drb/drb.rb:744:in `rescue in block in open'
```

Note, the drbunix:///var/folders... unix socket being reported on one side but
the druby://92.242... TCP socket reporting a timeout.  Weird.

Using verizon fios dns at home, somehow DRb resolves DRb.start_service to a remote
DRb service, which I believe happens below because we're not passing a URI so we
have no hostname from this URI and it tries to resolve the local hostname:
https://github.com/ruby/ruby/blob/v2_6_5/lib/drb/drb.rb#L879-L884

```
irb(main):001:0> s = DRb.start_service
irb(main):002:0> s.uri
=> "druby://92.242.140.21:51000"
```

Note, this is the IP address of verizon's DNS assistance program, as mentioned
here:
https://askubuntu.com/questions/587895/why-is-the-ip-92-242-140-21-connecting-to-one-of-my-ports-is-it-malware#comment1583793_587954

We can avoid this entirely by explicitly telling DRb to use a local URI for our
DRb client:

```
irb(main):020:0>   Dir::Tmpname.create("automation_engine_client", nil) do |path|
irb(main):021:1*     puts DRb.start_service("drbunix://#{path}").uri
irb(main):022:1>     FileUtils.chmod(0o750, path)
irb(main):023:1>   end; nil
drbunix:///var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-69238-im1kfc
```

We've seen this weird thing in the past where our server was using unix sockets
and the client was failing, trying to access a TCP DRb server.  The never merged
solution in the past was to remove the TCP socket option from the server but I
believe the problem might actually have been what you see above, in the DRb client,
which also starts a drb server to talk to the unix socket drb server.

https://github.com/ManageIQ/manageiq-automation_engine/pull/234

Thanks @agrare with helping me debug this.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -120,7 +120,16 @@ begin
   # undefine Object#display which would be called over service#display
   DRbObject.send(:undef_method, :display)
 
-  DRb.start_service
+  # DRb.start_service with no URI can attempt to resolve your local hostname[1], which:
+  #   * is slower than just telling it to use a local address/socket
+  #   * could be wrong and in some cases, it can be a remote IP to a DNS assistance program
+  # [1] https://github.com/ruby/ruby/blob/v2_6_5/lib/drb/drb.rb#L879-L884
+  require 'tmpdir'
+  Dir::Tmpname.create("automation_client", nil) do |path|
+    DRb.start_service("drbunix://\#{path}")
+    FileUtils.chmod(0o750, path)
+  end
+
   $evmdrb = DRbObject.new_with_uri(MIQ_URI)
   raise AutomateMethodException,"Cannot create DRbObject for uri=\#{MIQ_URI}" if $evmdrb.nil?
   $evm = $evmdrb.find(MIQ_ID)
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -42,7 +42,7 @@ module MiqAeEngine
       require 'tmpdir'
       Dir::Tmpname.create("automation_engine", nil) do |path|
         self.drb_server = DRb.start_service("drbunix://#{path}", drb_front, :idconv => global_id_conv)
-        FileUtils.chmod(0o750, path)
+        FileUtils.chmod(0o700, path)
       end
     end
 
@@ -127,7 +127,7 @@ begin
   require 'tmpdir'
   Dir::Tmpname.create("automation_client", nil) do |path|
     DRb.start_service("drbunix://\#{path}")
-    FileUtils.chmod(0o750, path)
+    FileUtils.chmod(0o700, path)
   end
 
   $evmdrb = DRbObject.new_with_uri(MIQ_URI)


jrafanie:
Maybe this should use a local unix socket?  Thoughts?

Fryguy:
@jrafanie As discussed, should we just use a unix socket instead?

Fryguy:
Also, FWIW, on my machine here's what I get...also using verizon fios dns at home...

```ruby
require "drb"
drb = DRb.start_service
drb.uri
# => "druby://jfrey-mac.fios-router.home:56755"
```

That resolves to a 192.168.1.### address.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -120,7 +120,7 @@ class AutomateMethodException < StandardError
   # undefine Object#display which would be called over service#display
   DRbObject.send(:undef_method, :display)
 
-  DRb.start_service
+  DRb.start_service("druby://localhost:0")
```
mkanoor: @jrafanie Can we use a net stat like command to see how many sockets are opened and by which processes. Since the client is running as a separate process we should be able to get the port it is using.
The other thing is if we are using pipes to communicate from engine (server) to client, shouldn't we use pipes when communicating from client to engine (server). Not sure if this is an an Out Of Band messaging scheme from client to server. How would the engine (server side) know which client to contact. When we launch a method we set the MIQ_URI to tell the client where the server is listening, but how does the client tell the engine where his server is listening.

jrafanie: yeah, I was wondering if we should just use unix sockets on the "client" side of the drb communication too.  I'll change it to do that.  

My understanding is that both our automate server and client both start drb servers for communication but only the client requests the server's drb object but both need drb servers started.  I tried without `start_service` and it fails.

jrafanie: On my machine it creates a TCP socket:

```
irb(main):001:0> r = DRb.start_service
=> #<DRb::DRbServer:0x00007f947fcc0a90 @config={:idconv=>#<DRb::DRbIdConv:0x00007f94801940d0>, :verbose=>false, :tcp_acl=>nil, :load_limit=>26214400, :argc_limit=>256, :safe_level=>0}, @protocol=#<DRb::DRbTCPSocket:0x00007f947fcc04a0 @uri="druby://92.242.140.21:54402", @socket=#<TCPServer:fd 37, AF_INET, 0.0.0.0, 54402>, @config={:tcp_original_host=>"", :idconv=>#<DRb::DRbIdConv:0x00007f94801940d0>, :verbose=>false, :tcp_acl=>nil, :load_limit=>26214400, :argc_limit=>256, :safe_level=>0, :tcp_port=>54402}, @acl=nil, @msg=#<DRb::DRbMessage:0x00007f947fcc0478 @load_limit=26214400, @argc_limit=256>, @shutdown_pipe_r=#<IO:fd 38>, @shutdown_pipe_w=#<IO:fd 39>>, @uri="druby://92.242.140.21:54402", @exported_uri=["druby://92.242.140.21:54402"], @front=nil, @idconv=#<DRb::DRbIdConv:0x00007f94801940d0>, @safe_level=0, @grp=#<ThreadGroup:0x00007f947fcc0388>, @thread=#<Thread:0x00007f947fcc0338@/Users/joerafaniello/.rubies/ruby-2.6.5/lib/ruby/2.6.0/drb/drb.rb:1504 sleep>>
irb(main):002:0> r.uri
=> "druby://92.242.140.21:54402"
irb(main):003:0> Process.pid
=> 69469

```

```
> lsof -i -P |grep 69469
ruby      69469 joerafaniello   32u  IPv4 0x9455ca3322cc180b      0t0  TCP localhost:54395->localhost:11211 (ESTABLISHED)
ruby      69469 joerafaniello   37u  IPv4 0x9455ca3322cc7013      0t0  TCP *:54402 (LISTEN)
```

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -120,7 +120,16 @@ class AutomateMethodException < StandardError
   # undefine Object#display which would be called over service#display
   DRbObject.send(:undef_method, :display)
 
-  DRb.start_service
+  # DRb.start_service with no URI can attempt to resolve your local hostname[1], which:
+  #   * is slower than just telling it to use a local address/socket
+  #   * could be wrong and in some cases, it can be a remote IP to a DNS assistance program
+  # [1] https://github.com/ruby/ruby/blob/v2_6_5/lib/drb/drb.rb#L879-L884
+  require 'tmpdir'
+  Dir::Tmpname.create("automation_engine_client", nil) do |path|
```
Fryguy: Can multiple clients get created simulatneously?  Does there need to be something to tell them apart?

jrafanie:
Ok, I've updated the code to use a unix socket and updated the description.  I've tested this by going to Services -> Catalog, choosing "Order" for a service.  Prior to this code change, I would get timeout errors. With this code change, I can get the dropdown populated.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
agrare: Since the client is initiating a connection to the server I don't think the server needs to find a specific client (it couldn't before this change either).  We just need to make sure they don't collide which is what Tmpname is doing for us.

jrafanie: Good eye.  I can put `"automation_engine_client_#{Thread.current.object_id}"`.  What do you think?

jrafanie:
```
irb(main):002:0> t = []; 3.times { t << Thread.new { puts Thread.current.object_id} }; t.join
=> "#<Thread:0x00007fc1b8231988@(irb):2 run>#<Thread:0x00007fc1b8231848@(irb):2 run>#<Thread:0x00007fc1b8231730@(irb):2 run>"
70234997361860
70234997361700
70234997361560
```

chessbyte: Will this approach work in podified?  /cc @carbonin

jrafanie: oh yeah, each call to `Dir::Tmpname.create` will give a different path already, so it should be fine.

```
irb(main):004:0> Dir::Tmpname.create("automation_engine_client", nil) { |path| puts path }
/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-3wl6hd
=> "/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-3wl6hd"
irb(main):005:0> Dir::Tmpname.create("automation_engine_client", nil) { |path| puts path }
/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-140g3ya
=> "/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-140g3ya"
irb(main):006:0> Dir::Tmpname.create("automation_engine_client", nil) { |path| puts path }
/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-1b1gyci
=> "/var/folders/bf/hh6xb_k15wbfzg4tlbkzch300000gn/T/automation_engine_client20200313-70700-1b1gyci"
```

Fryguy: @jrafanie Turns out Tmpname takes care of it by putting random junk on the end...I misinterpreted the second `nil` (which I think is not needed based on the doc?)

carbonin: > Will this approach work in podified?

Yeah, I believe all the automate Drb is local to the worker. As long as multiple workers are not accessing the same Drb server it will be fine.

NickLaMuro: Yeah, it defaults to `Dir.tmpdir` if `nil`.

That said, if you wanted put this name in `/var/www/miq/vmdb/tmp`, you could then do:

```ruby
Dir::Tmpname.create("automation_engine_client", Rails.root.join("tmp").to_s) { |dir| ... }
```

mkanoor:
@jrafanie 
I dont think we can use static names, since there could be multiple instances of the engine running on a single appliance, each one would need a dedicated connection name.

mkanoor:
@jrafanie Plus you have methods that can invoke other methods, each method would need a dedicated connection name.

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
jrafanie: So, are we all good with the current code?   Via, https://github.com/ManageIQ/manageiq-automation_engine/pull/431#discussion_r392345943 it creates random paths each time it's called so each automate "client" that starts a DRb server will get a different path.

jrafanie:
> @jrafanie Plus you have methods that can invoke other methods, each method would need a dedicated connection name.

@mkanoor see the in-line comment above, I think it addresses your concern

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
agrare: :+1:

jrafanie:
One second... I just got this while testing David's original issue:

```
[----] E, [2020-03-13T15:40:35.536582 #77390:3fdaf18cd3e8] ERROR -- : Method STDERR: The following error occurred during inline method preamble evaluation:
[----] E, [2020-03-13T15:40:35.536745 #77390:3fdaf18cd3e8] ERROR -- : Method STDERR: ArgumentError: too long unix socket path (111bytes given but 104bytes max)
```

mkanoor:
👍

jrafanie:
> One second... I just got this while testing David's original issue:
> 
> ```
> [----] E, [2020-03-13T15:40:35.536582 #77390:3fdaf18cd3e8] ERROR -- : Method STDERR: The following error occurred during inline method preamble evaluation:
> [----] E, [2020-03-13T15:40:35.536745 #77390:3fdaf18cd3e8] ERROR -- : Method STDERR: ArgumentError: too long unix socket path (111bytes given but 104bytes max)
> ```

Ok, I shrunk the name to `automation_client` to match the length of the existing `automation_engine` used for the server.

I tested this a few more times and haven't seen any errors.

jrafanie:
@miq-bot rm_label wip

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -120,7 +120,16 @@ class AutomateMethodException < StandardError
...
 
-  DRb.start_service
+  # DRb.start_service with no URI can attempt to resolve your local hostname[1], which:
+  #   * is slower than just telling it to use a local address/socket
+  #   * could be wrong and in some cases, it can be a remote IP to a DNS assistance program
+  # [1] https://github.com/ruby/ruby/blob/v2_6_5/lib/drb/drb.rb#L879-L884
+  require 'tmpdir'
+  Dir::Tmpname.create("automation_client", nil) do |path|
+    DRb.start_service("drbunix://\#{path}")
+    FileUtils.chmod(0o750, path)
```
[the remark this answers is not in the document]
agrare: > I'm pretty sure 0o is a python 3 thing to state it is octal.

Isn't it a ruby thing also?
>> 0o750
=> 488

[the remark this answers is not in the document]
NickLaMuro: @kbrock Pretty sure that `rubocop` requires the `0o700` version, so I think LJ was probably following that.

Not saying I necessarily agree with it, but most likely what you are seeing here.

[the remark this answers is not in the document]
jrafanie: haha, I just copied what I did previously in the same file for https://github.com/ManageIQ/manageiq-automation_engine/pull/201

I'm looking to see if there was a reason behind it now...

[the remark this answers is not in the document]
Fryguy: I like the `0o` prefix to go along with `0x` and `0b`...personal opinion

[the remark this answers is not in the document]
Fryguy: We used specifically 750 previously when we tested permissions for the broker DRb security issue and that was sufficient, but if 700 works also, then it's always better to lock down.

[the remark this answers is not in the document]
jrafanie: I'll try 0o700. I don't mind the notation either way.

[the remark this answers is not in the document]
jrafanie: Ok, I tested it and I still get service dialog dropdown values when I order the dialog.  I'm not sure what else to test.

jrafanie:
@miq-bot add_label jansa/yes?

gmcculloug:
> We can avoid this entirely by

@jrafanie I was thinking the right answer was going to be switching from verizon.  But I guess this approach works as well.

Dropped these changes on an appliance I used yesterday for provisioning Services and VMs.  I ran simultaneous service and vm provisioning as well as retirement.  Everything looks good.

Thanks.

simaishi:
Jansa backport details:

```
$ git log -1
commit 8026eaa398a345b10167619a4dfc138fde624013
Date:   Wed Mar 18 19:47:29 2020 -0400

    Merge pull request #431 from jrafanie/force_client_drb_connection_locally

    Explicitly pass a local URI for our client's DRb.start_service

    (cherry picked from commit 3b71688b4eff4f6ba2741168372cc14dbd541a98)
```

PR #442, opened 2020-05-13
WIP: Automate data models

kbrock:
These models are automate related and we are adding them over here.

This is the other side of https://github.com/ManageIQ/manageiq/pull/20167
cross repo tests: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/122


kbrock:
too much has changed to move this across.

Maybe at a future time we can do this work

PR #439, opened 2020-04-06
Improve automate query performance.

lfu:
Columns `domain_id`  and `relative_path` are added into` MiqAeNamespace`, `MiqAeClass`, `MiqAeInstance` and `MiqAeMethod`.

Depends on https://github.com/ManageIQ/manageiq/pull/20050
Depends on https://github.com/ManageIQ/manageiq-schema/pull/465

cross repo tests: ManageIQ/manageiq-cross_repo-tests#112

Fixes https://github.com/ManageIQ/manageiq-automation_engine/issues/410

PR #452, opened 2020-08-08
[WIP] prune down get_homonymic

kbrock:
Right now, this is opening a conversation with @lfu 

As far as I can tell, `get_homonymic` is basically `find_best_match_by`

- `get_homonymic` there are 2 very different signatures depending upon if you have a `relative_path` or a `fqname`.
- `find_best_match_by` doesn't check that the domains are enabled.
- `get_homonymic` typically doesn't check if the domain is enabled (expected the default to be to check).
- `get_homonymic` handles urls - not sure if this is a valid use case.
- `find_best_match_by` needs to be split out concepts of `visible_by_user()`, `domain_priority_order`, `best_match` (with a where), and `find_by` (use a `first`)
- `find_best_match_by` allows wild characters for some models. Do we really need this? Where is this actually valid?
- the lookup by user concept feels like it belongs in rbac and not here.

Could we consolidate this functionality and put into the models, `RelativePathMixin`, or `MiqAeDatastore`.

Since this is basically an enhanced `find_by`, and you need to pass `self.class` into the `get_homonymic` methods, I'm leaning towards moving it into `RelativePathMixin`

There are a few changes in the models, but they are basically consolidating to one method call.

Wanted to know where I should consolidate this stuff before I go all in with these changes.

There is a lot of code around homonymic and it is very inefficient the way it looks up domains and sorts the methods/instances/...

Usage:

- `get_homonymic` is used in 3 places in the ui, 1 of those calls being to the generic `klass`.
- `find_best_match_by is called in 1 place by `MiqAeDomainSearch` - the call uses a generic `klass`.

### DIFF
--- a/app/models/miq_ae_datastore.rb
+++ b/app/models/miq_ae_datastore.rb
@@ -230,36 +230,42 @@ module MiqAeDatastore
     MiqAeDomain.reset_priorities
   end
 
-  def self.get_homonymic_across_domains(user, arclass, fqname, enabled = nil)
+  # @param prefix (default true) the domain name is a prefix
+  def self.get_homonymic_across_domains(user, arclass, fqname, enabled = nil, prefix: true)
     return [] if fqname.blank?
 
-    options = arclass == ::MiqAeClass ? {:has_instance_name => false} : {}
-    _, ns, klass, name = ::MiqAeEngine::MiqAePath.get_domain_ns_klass_inst(fqname, options)
-    name = klass if arclass == ::MiqAeClass
-    MiqAeDatastore.get_sorted_matching_objects(user, arclass, ns, klass, name, enabled)
-  end
-
-  def self.get_sorted_matching_objects(user, arclass, namespace, klass, name, enabled)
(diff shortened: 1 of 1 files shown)


kbrock:
@Fryguy I'm guessing we want to close all automate performance efforts?

Most of the work I did was improving the lookup mechanism and getting us closer to passing around objects instead of the path strings.

background: methods accept a path string, use the path to lookup an object, do work, and then convert back to a path. Then the next method used the path or a child path, looks up an object, ... This can't really use includes or ancestry (even though domains are very ancestry friendly) and basically forces N+1's.

Lucy and I were making headway (e.g.: #439) mostly in the realm of deleting inconsistent lookup logic headed towards reducing lookups and passing objects. Having `Domain` in core and `Namespace` in this repo made this work tricky.

My knowledge is rusty, and I don't have Lucy as a resource.

Verifying: we want to just close these efforts

Fryguy:
Yeah if we're moving towards works let's not put effort into it

PR #454, opened 2020-08-13
Remove unused refresh_new_target builtin method

agrare:
The refresh_new_target method is deprecated and only has one user (RHV).  Once that is deleted we can remove this method.

Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-ovirt/pull/515
- [x] https://github.com/ManageIQ/manageiq-content/pull/666


agrare:
> Coverage decreased (-0.01%) to 85.739% 

Not sure how a pure-deletions PR can decrease spec coverage :laughing:

lfu:
How are [this method](https://github.com/ManageIQ/manageiq/blob/db5d90f8ec911f15aca98e445aedc6ab8a4b270d/app/models/ems_event/automate.rb#L25)? Is it still being used?

agrare:
After this PR that method is no longer used, see https://github.com/ManageIQ/manageiq/issues/20443

PR #455, opened 2020-09-10
Plugin Generator Updates

Fryguy:
Plugin Generator Updates

### DIFF
--- a/lib/tasks/README.md
+++ b/lib/tasks/README.md
@@ -1,6 +1,20 @@
-Tasks (.rake files) in this directory will be available in the main ManageIQ app.
-They can be executed in the provider gem via the app: namespace
+Tasks (.rake files) in this directory will be available as public tasks in the main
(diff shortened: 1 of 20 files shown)


Fryguy:
@gtanzillo Please review.  I am blasting out changes form the plugin generator to all repos.

lib/tasks/README.md
```diff
@@ -1,6 +1,20 @@
...
+```ruby
+namespace "manageiq:automation_engine" do
+  desc "Explaining what the task does"
+  task :your_task do
+    # Task goes here
+  end
+end
+```
+
+Tasks places in the lib/tasks_private directory will be private to the plugin
```
gtanzillo, line 19: Should be `Tasks placed ...`

Fryguy: I'll have to change that in the generator and blast it out separately. Let's leave it here for now. I have a couple of other things to blast out in the next round.

PR #460, opened 2020-10-27
rename factorygirl to factorybot

d-m-u:
there is the one of them left in this repo, should be factory bot


d-m-u:
@Fryguy said previous we might get away with ignoring this cop on https://github.com/ManageIQ/manageiq-schema/pull/492#issuecomment-715546573, might also apply here

PR #475, opened 2021-03-30
Fix issue where delivery failures return 'true'

Fryguy:
This method is expected to return nil on failures, otherwise the
workspace itself, however when there is an error, _log.error will return
true, which becomes the return value of this method.


chessbyte (commented):
Is the same `return nil` needed after line 118?

Fryguy:
@chessbyte No, because that will fall through to line 124 (`return_results`), which does the right thing.

chessbyte:
@Fryguy thanks - as a general observation, this method is too long.  In a future PR, would be nice to refactor it so we can see the forest for the trees.

lfu (approved):
LGTM

Fryguy:
Backported to `lasker` in commit 5eeb6aaf0106910ae5eeb5c80d95d3f68ed6e06a.

```text
commit 5eeb6aaf0106910ae5eeb5c80d95d3f68ed6e06a
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Wed Mar 31 08:58:15 2021 -0400

    Merge pull request #475 from Fryguy/fix_deliver_return
    
    Fix issue where delivery failures return 'true'
    
    (cherry picked from commit 5be34d7779dbd7e968326076c39aacca197fc59f)
```

PR #487, opened 2021-11-05
Fix tests for manageiq-password 1.0 upgrade

Fryguy:
This change, along with ManageIQ/manageiq-password#26 will handle the
changes for the 1.0 upgrade of manageiq-password

Depends on ManageIQ/manageiq-password#26 and subsequent patch release.

### DIFF
--- a/spec/models/miq_ae_password_spec.rb
+++ b/spec/models/miq_ae_password_spec.rb
@@ -26,8 +26,12 @@ describe MiqAePassword do
       expect(described_class.decrypt(ManageIQ::Password.encrypt(plaintext))).to eq(plaintext)
     end
 
-    it "throws understandable error" do
-      expect { described_class.decrypt("v1:{something}") }.to raise_error(ManageIQ::Password::PasswordError)
+    it "throws error for plaintext password" do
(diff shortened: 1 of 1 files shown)


spec/models/miq_ae_password_spec.rb
```diff
@@ -26,8 +26,12 @@
...
     end
 
-    it "throws understandable error" do
-      expect { described_class.decrypt("v1:{something}") }.to raise_error(ManageIQ::Password::PasswordError)
+    it "throws error for plaintext password" do
+      expect { described_class.decrypt("passw0rd") }.to raise_error(ManageIQ::Password::PasswordError)
+    end
+
+    it "throws error for undecryptable strings" do
+      expect { described_class.decrypt("v2:{something}") }.to raise_error(ManageIQ::Password::PasswordError)
```
Fryguy, line 34: The diff doesn't show it well, but this test changed from `v1` to `v2`, since with manageiq-password 1.0 v1 keys are no longer supported and thus appear like plaintext.  The spirit of the test is "undecryptable strings", hence the change.

agrare (approved):

PR #490, opened 2021-12-20
Expose VMware VM rename operation to automate

agrare:
Ref: https://talk.manageiq.org/t/renaming-a-vm-through-automate/5606/3


agrare:
NOTE I would have put this in the base VmOrTemplate service model but this is currently only implemented by VMware and automate doesn't have a good "check supports?(feature)" pattern (which we really should address)

PR #503, opened 2022-07-26
[wip]Adding logs for requests

akhilkr128:
For the [feature](https://github.com/ManageIQ/manageiq/issues/21188)
Work in progress

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -15,6 +15,7 @@ module MiqAeEngine
   end
 
   def self.deliver_queue(args, options = {})
+    # binding.pry
     options = {
       :class_name  => 'MiqAeEngine',
       :method_name => 'deliver',
@@ -67,6 +68,20 @@ module MiqAeEngine
     options[:ae_state_previous] = YAML.dump(workspace.current_state_info) unless workspace.current_state_info.empty?
   end
 
+  def self.get_request_id(options)
+    # binding.pry
+    if options[:object_type].ends_with?("Request")
+      return options[:object_id]
+    end
+    if options[:object_type].ends_with?("Task")
+      return MiqRequestTask.find(options[:object_id]).miq_request_id
+    end
+    if options[:object_type].starts_with?("Service")
+      return Service.find(264).miq_request.id
+    end
+    nil
+  end
+
   def self.deliver(*args)
     options     = options_from_args(args)
     user_obj    = ae_user_object(options)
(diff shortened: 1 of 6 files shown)


lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -67,45 +68,67 @@ def self.deliver_queue(args, options = {})
...
     options     = options_from_args(args)
     user_obj    = ae_user_object(options)
     state       = options[:state]
     vmdb_object = nil
     ae_result   = 'error'
     miq_task    = MiqTask.find(options[:open_url_task_id]) if options[:open_url_task_id]
-
+    $request_id = get_request_id(options)
+    $object_id = options[:object_id]
+    $object_type = options[:object_type]
```
Fryguy, line 94: We shouldn't be using global vars here...perhaps they can be stored in the workspace?

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -114,17 +137,19 @@ def self.deliver(*args)
         else
           if ae_result.casecmp('error').zero?
             miq_task&.update_message(MiqTask::MESSAGE_TASK_COMPLETED_UNSUCCESSFULLY)
-            message = "Error delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate: #{ws.root['ae_message']}"
+            message = "Error delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate: #{ws.root['ae_message']}"
             _log.error(message)
+            RequestLogs.create(:log_message => message, :miq_requests_id => $request_id, :object_id => $object_id, :object_type => $object_type)
```
Fryguy, line 142: I think we discussed this elsewhere, and I'm fine with this for now, but we might want to create a specialized RequestLogger that writes to the table as well as the log.

lib/miq_automation_engine/engine/miq_ae_engine.rb
akhilkr128: ok

lib/miq_automation_engine/engine/miq_ae_engine.rb
akhilkr128: Working on the logger.

Fryguy (changes requested):
https://github.com/ManageIQ/manageiq-automation_engine/pull/503/files#r934935652 needs to be resolved before this can be merged.  A global variable there is a no-go because then we can't have multiple resolutions simultaneously.  I'm not sure what the exact right approach should be...perhaps it can be stored in the workspace or the miq_task?

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -67,45 +68,67 @@ def self.deliver_queue(args, options = {})
...
+  def self.get_request_id(options)
+    # binding.pry
+    if options[:object_type].ends_with?("Request")
+      return options[:object_id]
+    end
+    if options[:object_type].ends_with?("Task")
+      return MiqRequestTask.find(options[:object_id]).miq_request_id
+    end
+    if options[:object_type].starts_with?("Service")
+      return Service.find(264).miq_request.id
```
Fryguy, line 80: This is hardcoding a specific service, so will need to be fixed before completion.

Fryguy:
Closing in favor of https://github.com/ManageIQ/manageiq-automation_engine/pull/516

PR #505, opened 2022-08-17
comment the supports feature code

kbrock:
We have `validate_resize` for backwards compatibility.
It feels safer to just implement this rather than ask the customer
to find this method and update it.

the 2 validate methods look like they are implementing the AvailabilityMixin
but they are just a boolean method instead

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
@@ -5,7 +5,9 @@ module MiqAeMethodService
     expose :resize_revert,  :override_return => nil
 
     expose :supports_resize?
+    # @return [boolean] (not using supports)
     expose :validate_resize_confirm
+    # @return [boolean] (not using supports)
     expose :validate_resize_revert
 
     expose :associate_floating_ip,    :override_return => nil
@@ -22,6 +24,9 @@ module MiqAeMethodService
       sync_or_async_ems_operation(options[:sync], "detach_volume", [volume_id])
     end
 
+    # this implents the AvailabilityMixin interface
+    # for backwards compatibility in customer scripts
+    # prefer using supports_resize? method instead
     def validate_resize
       {:available => supports_resize?, :message => unsupported_reason(:resize)}
     end


lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
```diff
@@ -22,6 +24,9 @@ def detach_volume(volume_id, options = {})
       sync_or_async_ems_operation(options[:sync], "detach_volume", [volume_id])
     end
 
+    # this implents the AvailabilityMixin interface
+    # for backwards compatibility in customer scripts
+    # prefer using supports_resize? method instead
```
bdunne, line 29: Should this also `Vmdb::Deprecation.deprecate_methods(self, :validate_resize => :supports_resize?)` or something similar?

lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
```diff
@@ -22,6 +24,9 @@ def detach_volume(volume_id, options = {})
...
     def validate_resize
       {:available => supports_resize?, :message => unsupported_reason(:resize)}
```
Fryguy, line 31: Do we want to add an ActiveSupport::Deprecation here or something?

kbrock:
@bdunne @Fryguy I mostly wanted to document it as we were greping our code base.


both versions of this method have been around since 2016.
Since these methods are only used by customer automate scripts running in production, I thought it didn't make sense to introduce the deprecation messages.

I'm also not sure if we want to push `supports_resize` or `supports?(:resize)`, or just let them all slide.

refs:
- verify https://github.com/ManageIQ/manageiq/pull/7922
- supports https://github.com/ManageIQ/manageiq/pull/9683

kbrock:
update:

- rebased

I still feel it is best to just add this as a reference for our future selves and not deprecate or make any changes here.

Fryguy:
> Since these methods are only used by customer automate scripts running in production, I thought it didn't make sense to introduce the deprecation messages.

That's exactly why we want to mark it deprecated - at least they should see the deprecation in the logs, and then we can remove it later.

PR #509, opened 2022-09-06
Support ruby 3

jrafanie:
* rspec-mocks 3.10.3 requires "with" to be explicit with hashes vs. kwargs
* Handle ruby 2.6-3.0+ kwargs/options
  Change `execute_with_user` to capture any kwargs passed as a hash and append to args passed in and call the `MiqAeServiceMethods` method with these args but never send over the kwargs.  Therefore all `MiqAeServiceMethods` methods can be made to only accept options hashes and never kwargs.  This works with ruby 2.6-3.0+. We can change this when we drop ruby 2.
* Add tests for methods to exercise ruby 3/2 args
* Fix bug where conditional is always true, and handling nil/empty array

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
@@ -36,7 +36,7 @@ module MiqAeEngine
     end
 
     def self.miq_send_email(_obj, inputs)
-      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], :cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"])
+      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], {:cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"]})
     end
 
     def self.miq_snmp_trap_v1(_obj, inputs)
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
@@ -36,7 +36,7 @@ module MiqAeEngine
     end
 
     def self.miq_send_email(_obj, inputs)
-      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], {:cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"]})
+      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], :cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"])
     end
 
     def self.miq_snmp_trap_v1(_obj, inputs)
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
@@ -201,14 +201,20 @@ module MiqAeMethodService
     def ldap
     end
 
-    def execute(method_name, *args)
-      User.with_user(@workspace.ae_user) { execute_with_user(method_name, *args) }
+    def execute(method_name, *args, **kwargs, &block)
+      User.with_user(@workspace.ae_user) { execute_with_user(method_name, *args, **kwargs, &block) }
     end
 
-    def execute_with_user(method_name, *args)
+    def execute_with_user(method_name, *args, **kwargs, &block)
       # Since each request from DRb client could run in a separate thread
       # We have to set the current_user in every thread.
-      MiqAeServiceMethods.send(method_name, *args)
+
+      # For ruby 2.6-3.0+ support, we grab any kwargs and append as a hash
+      # at the end of the args, as all MiqAeServiceMethods now don't accept
+      # kwargs.  When we get to ruby 3, we can remove this and convert all to
+      # kwargs.
+      args << kwargs unless kwargs.blank?
+      MiqAeServiceMethods.send(method_name, *args, &block)
     rescue NoMethodError => err
       raise MiqAeException::MethodNotFound, err.message
     end
--- a/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
@@ -7,18 +7,29 @@ module MiqAeMethodService
 
     SYNCHRONOUS = Rails.env.test?
 
-    def self.send_email(to, from, subject, body, options = {}) # rubocop:disable Naming/MethodParameterName
+    def self.send_email(to, from, subject, body, *args, **kwargs)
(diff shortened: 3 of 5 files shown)


jrafanie:
cc @Fryguy do you have opinions about using the `ruby2_keywords` for the very open ended `execute` methods that users can be using with old style interfaces.  It requires we drop ruby 2.6 here but I think that's what we're doing anyway.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -201,11 +201,11 @@ def datastore
     def ldap
     end
 
-    def execute(method_name, *args)
+    ruby2_keywords def execute(method_name, *args)
```
jrafanie, line 204: TODO:  try removing ruby2_keywords and see if I can change options hash methods in `MiqAeServiceMethods` to use kwargs and use `**` in caller delegates such as this.

```
(method_name, *args, **kwargs)
```

jrafanie: I've dropped ruby2_keywords and found a temporary solution I'm fine with.  We basically accept either kwargs/options hashes coming into the execute methods but call the underlying method with only args appended with a hash of kwargs if kwargs exist.

I had to change one method to do no longer accept kwargs and instead pass an options hash.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -7,16 +7,16 @@ class MiqAeServiceMethods
 
     SYNCHRONOUS = Rails.env.test?
 
-    def self.send_email(to, from, subject, body, content_type: nil, cc: nil, bcc: nil) # rubocop:disable Naming/MethodParameterName
+    def self.send_email(to, from, subject, body, options = {}) # rubocop:disable Naming/MethodParameterName
```
jrafanie, line 10: This method needed to be changed to no longer accept kwargs as the caller in `execute_with_user` will append the kwargs to the incoming arguments as a hash so no kwargs will be sent.  I also need to change `miq_send_email` to no longer send kwargs as it bypassed `execute_with_user`.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -144,7 +144,7 @@ def self.create_provision_request(*args)
       # Need to add the username into the array of params
       # TODO: This code should pass a real username, similar to how the web-service
       #      passes the name of the user that logged into the web-service.
-      args.insert(1, User.lookup_by_userid("admin")) if args.kind_of?(Array)
+      args << User.lookup_by_userid("admin")
```
jrafanie, line 147: This fixes two bugs:
1) args is always an array because it's splatted
2) it inserted the user as the second element in the array, so, if you had an empty array, it would return `[nil, user]` on this line.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb
```diff
@@ -201,14 +201,20 @@ def datastore
...
       # Since each request from DRb client could run in a separate thread
       # We have to set the current_user in every thread.
-      MiqAeServiceMethods.send(method_name, *args)
+
+      # For ruby 2.6-3.0+ support, we grab any kwargs and append as a hash
+      # at the end of the args, as all MiqAeServiceMethods now don't accept
+      # kwargs.  When we get to ruby 3, we can remove this and convert all to
+      # kwargs.
+      args << kwargs unless kwargs.blank?
+      MiqAeServiceMethods.send(method_name, *args, &block)
```
jrafanie, line 217: This is really the biggest change.  We capture the kwargs from ruby 2.7/3 and pass them along as a hash to the end of the args.  Then, we can keep all of the existing methods as accepting options hashes except for the one that was accepting kwargs, which I change below accept hashes instead.

There isn't another way to handle the situation in which existing automate code is calling execute with `:a =>` at that will be kwargs on ruby 3 even if we expect hashes.

Once we're on ruby 3+ only, we can use `**kwargs` and always pass along kwargs and all the methods can be made to accept hashes as kwargs.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
```diff
@@ -36,7 +36,7 @@ def self.miq_log_workspace(obj, _inputs)
     end
 
     def self.miq_send_email(_obj, inputs)
-      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], :cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"])
+      MiqAeMethodService::MiqAeServiceMethods.send_email(inputs["to"], inputs["from"], inputs["subject"], inputs["body"], {:cc => inputs["cc"], :bcc => inputs["bcc"], :content_type => inputs["content_type"]})
```
jrafanie, line 39: I needed to change this as the callee's interface was changed to support ruby 2.6-3.0+ so it no longer accepts keyword arguments directly.  Notice, this is because `execute_with_user` below now captures any kwargs as a hash and appends that to the args so no kwargs are sent to the callee in `MiqAeServiceMethods`.

kbrock: A little surprised that you needed this change.
Seems it would have been converted to a hash automatically in ruby

jrafanie: On ruby 2/3, the `send_email` interface is 5 arguments via this PR.  This call was passing only 4 on ruby 3 as kwargs are not an "argument".

jrafanie: I can try to avoid changing the interface of `send_email` and see if I can "fix" the inputs to work with ruby 2/3.  I thought I tried it but now that there are tests, we can experiment.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
jrafanie: This is now stale, I figured out a way to treat this method differently and accept kwargs or options hashes so we don't need to break existing callers using kwargs.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
jrafanie: This is now stale.  I figured out a way to fix send_email to accept kwargs or an options hash and treat it differently since it's really the only one doing kwargs.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb
```diff
@@ -7,18 +7,29 @@ class MiqAeServiceMethods
 
     SYNCHRONOUS = Rails.env.test?
 
-    def self.send_email(to, from, subject, body, content_type: nil, cc: nil, bcc: nil) # rubocop:disable Naming/MethodParameterName
+    def self.send_email(to, from, subject, body, *args, **kwargs)
```
jrafanie, line 10: This method now accepts being called from `execute_with_user`, which will take the kwargs and append them to the args and send them along here.  Additionally, it accepts kwargs as it previously did by using any kwargs that are provided.

jrafanie: Technically, this changes the interface of this method but doesn't break callers which used the original kwargs interface.

PR #514, opened 2022-11-21
Deploy server profile template

eilam20:
Added this function for use in the automation engine.

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
@@ -5,8 +5,11 @@ module MiqAeMethodService
     expose :power_on
     expose :power_off
 
-    def self.create_server_profile_and_deploy_task(template_id, server_id, profile_name)
-      PhysicalServerProfileTemplate.deploy_server_from_template(template_id, server_id, profile_name)
+    def self.create_server_profile_and_deploy_task(ems_id, template_id, server_id, profile_name)
+      ext_management_system = ExtManagementSystem.find(ems_id)
+      profile_template = PhysicalServerProfileTemplate.find(template_id)
+      task_id = profile_template.deploy_server_from_template_queue(ext_management_system, server_id, profile_name)
+      MiqTask.wait_for_taskid(task_id)
     end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
```diff
@@ -4,5 +4,9 @@ class MiqAeServicePhysicalServer < MiqAeServiceModelBase
     expose :turn_off_loc_led
     expose :power_on
     expose :power_off
+
+    def self.create_server_profile_and_deploy_task(template_id, server_id, profile_name)
+      PhysicalServerProfileTemplate.deploy_server_from_template(template_id, server_id, profile_name)
```
Fryguy, line 9: I don't think this is right because this will run on the client side as opposed to the server side.  @agrare Can you comment here?  Maybe I'm wrong but I thought we did these different.

eilam20: @Fryguy what is the right way to do this? we need to call cisco intersight API client functions in the manageiq automation engine and deploy a physical server with the given server profile template

lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
```diff
@@ -4,5 +4,12 @@ class MiqAeServicePhysicalServer < MiqAeServiceModelBase
     expose :turn_off_loc_led
     expose :power_on
     expose :power_off
+
+    def self.create_server_profile_and_deploy_task(ems_id, template_id, server_id, profile_name)
+      ext_management_system = ExtManagementSystem.find(ems_id)
```
agrare, line 9: Hm according to https://github.com/ManageIQ/manageiq/pull/22241/files#diff-8d286308b75ae7dc8a266753732b0c1709d33293518c479765e34cadf8207910R90 a profile_template belongs to an ext_management_system so this shouldn't be required.  If you need access to the EMS you can reference it from the `profile_template` instance

lib/miq_automation_engine/service_models/miq_ae_service_physical_server.rb
eilam20: fixed

agrare (approved):

agrare:
NOTE this depends on https://github.com/ManageIQ/manageiq/pull/22241 which introduces the `deploy_server_from_template_queue` method

PR #516, opened 2023-01-06
Add logging to RequestLog model

DavidResende0:
Parent Issue: https://github.com/ManageIQ/manageiq/issues/21188
Continuation of: https://github.com/ManageIQ/manageiq-automation_engine/pull/503

### DIFF
--- a/lib/manageiq/automation_engine/logger.rb
+++ b/lib/manageiq/automation_engine/logger.rb
@@ -1,0 +1,52 @@
+module ManageIQ
+  module AutomationEngine
+    class Logger < ManageIQ::Loggers::Base
+      def add(severity, message = nil, progname = nil, resource_id: nil)
+        # Copied from Logger#add
+        severity ||= UNKNOWN
+        if @logdev.nil? or severity < level
+          return true
+        end
+        if progname.nil?
+          progname = @progname
+        end
+        if message.nil?
+          if block_given?
+            message = yield
+          else
+            message = progname
+            progname = @progname
+          end
+        end
+
+        RequestLog.create(:message => message, :severity => format_severity(severity), :resource_id => resource_id) if resource_id
+
+        super(severity, message, progname)
+      end
+
+      def info(progname = nil, resource_id: nil, &block)
+        add(INFO, nil, progname, resource_id: resource_id, &block)
+      end
+
+      def debug(progname = nil, resource_id: nil, &block)
+        add(DEBUG, nil, progname, resource_id: resource_id, &block)
+      end
+
+      def warn(progname = nil, resource_id: nil, &block)
+        add(WARN, nil, progname, resource_id: resource_id, &block)
+      end
+
+      def error(progname = nil, resource_id: nil, &block)
+        add(ERROR, nil, progname, resource_id: resource_id, &block)
+      end
+
+      def fatal(progname = nil, resource_id: nil, &block)
+        add(FATAL, nil, progname, resource_id: resource_id, &block)
+      end
+
+      def unknown(progname = nil, resource_id: nil, &block)
+        add(UNKNOWN, nil, progname, resource_id: resource_id, &block)
+      end
+    end
+  end
+end
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -67,6 +67,11 @@ module MiqAeEngine
     options[:ae_state_previous] = YAML.dump(workspace.current_state_info) unless workspace.current_state_info.empty?
   end
 
+  # Helper method to log the request to both the request_logs table and $log if a request_id is present
+  private_class_method def self.request_log(severity, message = nil, resource_id: nil, &block)
+    MiqRequest.request_log(severity, message, resource_id: resource_id, &block)
+  end
+
   def self.deliver(*args)
     options     = options_from_args(args)
     user_obj    = ae_user_object(options)
@@ -78,7 +83,6 @@ module MiqAeEngine
     begin
       miq_task&.state_active
       object_name = "#{options[:object_type]}.#{options[:object_id]}"
-      _log.info("Delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate")
       automate_attrs = automate_attrs_from_options(options)
 
       if options[:object_type]
@@ -89,11 +93,21 @@ module MiqAeEngine
       end
 
       uri = create_automation_object(options[:instance_name], automate_attrs, create_automation_object_options(options, vmdb_object))
+
+      if !uri.nil?
+        _scheme, _userinfo, _host, _port, _registry, _path, _opaque, query, _fragment = MiqAeUri.split(uri, "miqaedb")
+        args = MiqAeUri.query2hash(query)
+        miq_request_id = MiqAeWorkspaceRuntime.find_miq_request_id(args)
+      end
+
+      attr_string = ManageIQ::Password.sanitize_string(options[:attrs].inspect)
+      request_log(:info, "Delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate", :resource_id => miq_request_id)
+
       ws  = resolve_automation_object(uri, user_obj)
 
       if ws.nil? || ws.root.nil?
-        message = "Error delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate: Empty Workspace"
-        _log.error(message)
+        message = "Error delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate: Empty Workspace"
+        request_log(:error, message, :resource_id => miq_request_id)
         return nil
       end
 
@@ -106,7 +120,7 @@ module MiqAeEngine
           change_options_by_ws(options, ws)
 
           message = "Requeuing #{ManageIQ::Password.sanitize_string(options.inspect)} for object [#{object_name}] with state [#{options[:state]}] to Automate for delivery in [#{ae_retry_interval}] seconds"
-          _log.info(message)
+          request_log(:info, message, :resource_id => miq_request_id)
           queue_options = {:deliver_on => deliver_on}
           queue_options[:server_guid] = MiqServer.my_guid if ws.root['ae_retry_server_affinity']
           miq_task&.state_queued
@@ -114,8 +128,8 @@ module MiqAeEngine
         else
           if ae_result.casecmp('error').zero?
             miq_task&.update_message(MiqTask::MESSAGE_TASK_COMPLETED_UNSUCCESSFULLY)
-            message = "Error delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate: #{ws.root['ae_message']}"
-            _log.error(message)
+            message = "Error delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate: #{ws.root['ae_message']}"
+            request_log(:error, message, :resource_id => miq_request_id)
           end
           MiqAeEvent.process_result(ae_result, automate_attrs) if options[:instance_name].to_s.casecmp('EVENT').zero?
         end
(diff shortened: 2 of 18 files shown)


DavidResende0:
@miq-bot assign @Fryguy

Fryguy:
Discussed with @DavidResende0 and we are skipping the bot suggestion because the current code stays consistent with the other trivial accessors in the file.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -78,54 +78,62 @@ def self.deliver(*args)
...
         automate_attrs[create_automation_attribute_key(vmdb_object)] = options[:object_id]
         vmdb_object.before_ae_starts(options) if vmdb_object.respond_to?(:before_ae_starts)
         vmdb_object.mark_execution_servers if vmdb_object.respond_to?(:mark_execution_servers)
       end
-
       uri = create_automation_object(options[:instance_name], automate_attrs, create_automation_object_options(options, vmdb_object))
       ws  = resolve_automation_object(uri, user_obj)
+      miq_request_id = nil
+
+      if !defined?(ws.root.attributes).nil?
```
Fryguy, line 92: not sure if `defined?` is the correct choice here - what is this trying to check?

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -78,54 +78,62 @@ def self.deliver(*args)
...
+      $request_log.request_info("Delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate", {:resource_id => miq_request_id})
 
       if ws.nil? || ws.root.nil?
-        message = "Error delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate: Empty Workspace"
-        _log.error(message)
+        message = "Error delivering #{attr_string} for object [#{object_name}] with state [#{state}] to Automate: Empty Workspace"
+        $request_log.request_error(message, {:resource_id => miq_request_id})
         return nil
       end
-
```
Fryguy, line 99: Why were these blank lines removed? Thinking we should leave the original code as untouched as possible.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb
```diff
@@ -19,6 +19,18 @@ class MiqAeBuiltinMethod
     SERVICE        = 'service'.freeze
     UNKNOWN        = 'unknown'.freeze
 
+    def self.find_miq_request_id(workspace)
```
Fryguy, line 22: I'm thinking this should be a method on the MiqAeWorkspace itself and that would DRY up the code.  Then below you could call `obj.workspace.find_miq_request_id`.  Similarly you could do that up above in MiqAeEngine.deliver, you could do `ws.find_miq_request_id`.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -6,6 +6,18 @@ def self.ae_methods_dir
       end
     end
 
+    def self.find_miq_request_id(workspace)
```
Fryguy, line 9: Same here.

lib/miq_automation_engine/engine/miq_ae_engine.rb
DavidResende0: Just wants to make sure `ws.root.attributes` exists so that it doesn't run into problems trying to grab `ws.root.attributes['miq_request_id']` for example

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -218,10 +233,14 @@ def self.run_method(cmd)
           yield stdin if block_given?
           stdin.close
           threads << Thread.new do
-            stdout.each_line { |msg| $miq_ae_logger.info("Method STDOUT: #{msg.strip}") }
+            stdout.each_line do |msg|
+              $request_log.request_info("Method STDOUT: #{msg.strip}", {:resource_id => miq_request_id}, false)
```
Fryguy, line 237: Hmm - this didn't come out the way I'd expect and I think I led you down the wrong path.

I _think_ maybe instead of RequestLogger, we should have modified the AutomateLogger to accept a request_id.  Then the AutomateLogger would additionally write to the Request table (or a simpler RequestLogger).

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -52,13 +52,43 @@ def self.instantiate_with_user(uri, user, attrs)
       workspace = MiqAeWorkspaceRuntime.new(attrs)
       self.current = workspace
       workspace.instantiate(uri, user, nil)
+
+      if !uri.nil?
+        _scheme, _userinfo, _host, _port, _registry, _path, _opaque, query, _fragment = MiqAeUri.split(uri, "miqaedb")
+        args = MiqAeUri.query2hash(query)
+        miq_request_id = find_miq_request_id(args)
```
Fryguy, line 59: At this point, you should have a workspace - can we pull the request id out of the workspace similar to how we do it everywhere else?

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -102,34 +132,33 @@ def varset(uri, value)
...
-
       raise MiqAeException::InvalidPathFormat, "Unsupported Scheme [#{scheme}]" unless MiqAeUri.scheme_supported?(scheme)
       raise MiqAeException::InvalidPathFormat, "Invalid URI <#{uri}>" if path.nil?
 
       message = fragment.blank? ? "create" : fragment.downcase
       args = MiqAeUri.query2hash(query)
+      miq_request_id = self.class.find_miq_request_id(args)
+
+      @dom_search.miq_request_id = miq_request_id
+      $request_log.request_info("Instantiating [#{ManageIQ::Password.sanitize_string(uri)}]", {:resource_id => miq_request_id}, false) if root.nil?
```
Fryguy, line 146: Though this one is trickier

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
DavidResende0: I originally had it that way but some spec tests failed because the workspace wasn't defined

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -102,34 +132,33 @@ def varset(uri, value)
     end
 
     def instantiate(uri, user, root = nil)
```
Fryguy, line 134: Do we know what is the caller of this?

lib/miq_automation_engine/engine/miq_ae_event.rb
```diff
@@ -67,6 +67,7 @@ def self.eval_alert_expression(target, inputs, options = {})
   def self.build_evm_event(event, passed_inputs = {})
     inputs = passed_inputs.dup
 
+    # To Do: Add to Request Logs
```
Fryguy, line 70: Use keywords (they show up in IDEs like VSCode and Sublime)

```suggestion
    # TODO: Add to Request Logs
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
DavidResende0: `self.instantiate_with_user` above

lib/miq_automation_engine/engine/miq_ae_engine.rb
Fryguy: Yeah defined is not the right way to do this... Think the following will work

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -78,7 +78,6 @@ def self.deliver(*args)
     begin
       miq_task&.state_active
       object_name = "#{options[:object_type]}.#{options[:object_id]}"
-      _log.info("Delivering #{ManageIQ::Password.sanitize_string(options[:attrs].inspect)} for object [#{object_name}] with state [#{state}] to Automate")
```
Fryguy, line 81: I'm not sure about this one - I'm concerned it's going to be logged too late if we do it after `resolve_automation_object`

lib/miq_automation_engine/engine/miq_ae_event.rb
```diff
@@ -122,14 +123,13 @@ def self.process_result(ae_result, aevent)
         inputs[:host]                  = Host.find_by(:id => aevent[:host_id])               unless aevent[:host_id].nil?
         inputs[:ext_management_system] = ExtManagementSystem.find_by(:id => aevent[:ems_id]) unless aevent[:ems_id].nil?
       end
-
       target     = inputs.delete(:target) || inputs['vm']
       event_name = args['logical_event'] || aevent[:event_type]
-      $miq_ae_logger.info("Enforcing Policy [#{ae_result}]")
+      $miq_ae_logger.info("Enforcing Policy [#{ae_result}]", :resource_id => miq_request_id)
```
Fryguy, line 128: I don't think this one needs to change, because policies shouldn't have requests associated.

lib/miq_automation_engine/engine/miq_ae_event.rb
```diff
@@ -122,14 +123,13 @@ def self.process_result(ae_result, aevent)
...
       MiqPolicy.enforce_policy(target, event_name, inputs) unless target.nil?
     end
   rescue URI::InvalidURIError => err
-    $miq_ae_logger.error(err.message)
+    $miq_ae_logger.error(err.message, :resource_id => miq_request_id)
```
Fryguy, line 132: Same here

Fryguy:
@DavidResende0 I updated a bunch of stuff here - Going to look at the core changes and regular `$log` changes tomorrow, but I think I might keep the `$request_log`, but instead make it just delegate to $log.

lib/miq_automation_engine/engine/miq_ae_engine.rb
Fryguy: We my want to keep this line as "Attempting to Deliver".  Then later between the create_automation_object and resolve_automation_object, we can try to get the request_id from the uri.

lib/manageiq/automation_engine/logger.rb
```diff
@@ -0,0 +1,52 @@
...
+          progname = @progname
+        end
+        if message.nil?
+          if block_given?
+            message = yield
+          else
+            message = progname
+            progname = @progname
+          end
+        end
```
Fryguy, line 20: I wish that was extracted in upstream Ruby - I'm considering pull this out into a private method upstream.

Fryguy: Actually, what they're probably going to suggest is to create a different LogDevice.  I might consider that.

Fryguy:
@DavidResende0 I'm seeing some issues locally caused by https://github.com/ManageIQ/manageiq-automation_engine/pull/519, but those are unrelated.  Hoping to get that merged first, then I can rebase this.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -89,11 +93,35 @@ def self.deliver(*args)
       end
 
       uri = create_automation_object(options[:instance_name], automate_attrs, create_automation_object_options(options, vmdb_object))
+      _scheme, _userinfo, _host, _port, _registry, _path, _opaque, query, _fragment = MiqAeUri.split(uri, "miqaedb")
+      args = MiqAeUri.query2hash(query)
+
+      miq_request_id = nil
```
Fryguy, line 99: @DavidResende0 Can you reuse `MiqAeWorkspaceRuntime.find_miq_request_id` here?  The one that uses args.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -52,13 +52,56 @@ def self.instantiate_with_user(uri, user, attrs)
...
+      $miq_ae_logger.error(err.message, :resource_id => miq_request_id)
     ensure
       clear_stored_workspace
     end
 
+    def self.find_miq_request_id(args)
+      miq_request_id = nil
+      if args['MiqRequest::miq_request']
+        miq_request_id = args['MiqRequest::miq_request']
+      elsif args['vmdb_object_type'].to_s.include?("task")
```
jrafanie, line 75: Can you verify the vmdb_object_type is lowercase all the time?

jrafanie (approved):
LGTM, let's tackle any changes in a followup.

(the defensive rescue on failing to extract the resource_id comes to mind)

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
DavidResende0: yeah just checked and can confirm that `vmdb_object_type` is always lowercase. It always looks something like `vm_retire_task`, `vm_retire_request`, or `vm`

PR #517, opened 2023-01-30
[WIP] Add fake interface for workflow execution, temporarily

Fryguy:
cc @agrare

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -69,6 +69,19 @@ module MiqAeEngine
 
   def self.deliver(*args)
     options     = options_from_args(args)
+
+    # HACK: If we are launching a specific Automate instance representing a workflow, then we run that instead
+    if options[:namespace] == "System" && options[:class_name] == "Workflow"
+      workflow_id = options[:instance_name].split("-").last.to_i
+      workflow = Workflow.find(workflow_id)
+
+      task_id = workflow.execute(inputs: options)
+      MiqTask.wait_for_taskid(task_id)
+
+      workflow_instance_id = MiqTask.find(task_id).context_data[:workflow_instance_id]
+      return WorkflowInstance.find(workflow_instance_id)
+    end
+
     user_obj    = ae_user_object(options)
     state       = options[:state]
     vmdb_object = nil


lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -69,6 +69,19 @@ def self.deliver_queue(args, options = {})
 
   def self.deliver(*args)
     options     = options_from_args(args)
+
+    # HACK: If we are launching a specific Automate instance representing a workflow, then we run that instead
```
kbrock, line 73: add TODO into this statement? (or does HACK show up in scans)

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -69,6 +69,19 @@ def self.deliver_queue(args, options = {})
...
+    if options[:namespace] == "System" && options[:class_name] == "Workflow"
+      workflow_id = options[:instance_name].split("-").last.to_i
+      workflow = Workflow.find(workflow_id)
+
+      task_id = workflow.execute(inputs: options)
```
kbrock, line 78: Huh. never thought I'd see the day that Jason would use JSON syntax over hash rockets

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -69,6 +69,19 @@ def self.deliver_queue(args, options = {})
...
+      MiqTask.wait_for_taskid(task_id)
```
kbrock, line 79: this wait_for_taskid makes me a little nervous here.
But if we need this temporarily then 👍

kbrock (approved):
Seems good for now.

Again, that wait in here really stands out since the others seem to exit with putting work into the queue.

lib/miq_automation_engine/engine/miq_ae_engine.rb
Fryguy: I've always said that this format makes sense for _keyword parameters_ which is what the syntax was designed for.  I'm against it for regular Hashes. IMO, Ruby should have made a distinction between Hashes and kwargs because they are different things. 😁

agrare:
> Again, that wait in here really stands out since the others seem to exit with putting work into the queue.

The normal automate deliver is blocking as well so the wait_for_task works the same way

Fryguy:
@agrare What I think we _should_ do, to be honest, is have a new Automation entrypoint that is plugin-agnostic.  This way we can start moving away from Automate to a generic interface.

agrare:
Yeah absolutely, if "automate" is an EmbeddedAutomationManager we could standardize on that.  Would also be interesting to see if we could run automate methods as a `Resource` in an ASL state machine

PR #521, opened 2023-02-15
Fix logger when wrapped in a container/journald logger

Fryguy:
The wrapped container/journald logger from core uses ActiveSupport::Logger.broadcast and then changes the underlying logger to have a nil logdev. This has the effect of writing to the wrapped logger and avoiding writing to the underlying file. However, in the case of this automation logger, we still want to write to the database, so we still need that underlying logger to execute the database part, but not the file part.

There is an added complication that ActiveSupport::Logger.broadcast fronts the `add` method, but doesn't support kwargs. Prior to this commit, we used kwargs to pass around the request_id, however, the broadcast would ignore those, instead sending them as a fourth paramater to the container loggers add method, which doesn't known about that extra parameter. To fix that issue, this commit instead creates a new method responsible for writing to the db, and then info, debug, etc will call that first before calling the original add method with the expected number of parameters.

@jrafanie Please review.

### DIFF
--- a/lib/manageiq/automation_engine/logger.rb
+++ b/lib/manageiq/automation_engine/logger.rb
@@ -1,11 +1,13 @@
 module ManageIQ
   module AutomationEngine
     class Logger < ManageIQ::Loggers::Base
-      def add(severity, message = nil, progname = nil, resource_id: nil)
-        # Copied from Logger#add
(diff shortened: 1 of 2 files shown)


jrafanie (approved):
LGTM

PR #537, opened 2023-12-04
permit_concurrent_loads is not needed with zeitwerk autoloader

jrafanie:
This was only needed for classic autoloader.  The core "freedom" patch made zeitwerk autoloader bypass the interlock anyway, so now that we're only supporting zeitwerk, this is no longer needed.

Co-dependency:
https://github.com/ManageIQ/manageiq/pull/22801

Part of the rails 7 upgrade: https://github.com/ManageIQ/manageiq/issues/22052


jrafanie:
This is ready to go since we were already defaulting to [zeitwerk on quinteros](https://github.com/ManageIQ/manageiq/blob/quinteros/config/application.rb#L99)

PR #538, opened 2023-12-18
Allow StateVarHash to YAML.load from runtime instantiate

jrafanie:
Previously, we allowed it to be permitted in serialized columns, but this test also requires it be permitted from miq_ae_workspace_runtime.rb instantiate.

Like https://github.com/ManageIQ/manageiq-providers-vmware/pull/892, this depends on:

- [ ] https://github.com/ManageIQ/manageiq/pull/22698
- [ ] https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/835


agrare (approved):

agrare:
Cross repo is green https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/835 merging on red

PR #542, opened 2024-02-29
drop supports_ methods

kbrock:
part of:
- [ ] https://github.com/ManageIQ/manageiq/pull/22883

Goal is to remove exposed methods that no longer exist.
Defining them where possible to keep backwards compatibility

- drop v2v vm_import (no longer a thing)
- backwards support for supports_resize

### DIFF
--- a/lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
+++ b/lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
@@ -4,7 +4,6 @@ module MiqAeMethodService
     expose :resize_confirm, :override_return => nil
     expose :resize_revert,  :override_return => nil
 
-    expose :supports_resize?
     # @return [boolean] (not using supports)
     expose :validate_resize_confirm
     # @return [boolean] (not using supports)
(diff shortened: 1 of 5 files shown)


lib/miq_automation_engine/service_models/miq_ae_service_manageiq-providers-openstack-cloud_manager-vm.rb
```diff
@@ -4,7 +4,6 @@ class MiqAeServiceManageIQ_Providers_Openstack_CloudManager_Vm < MiqAeServiceVmC
     expose :resize_confirm, :override_return => nil
     expose :resize_revert,  :override_return => nil
 
-    expose :supports_resize?
```
agrare, line 7: This is exposed so that users can call it from automate scripts, so I don't think we should remove this.

kbrock: agreed, this line says "please provide `Vm#supports_resize?`.

We have 2 options
- expose delegates over to Vm
- define a method here that calls over to vm.

We transitioned from the first option to the second option

kbrock: think maybe we need to use `object_send` here. researching

kbrock: fixed. thanks

agrare (approved):

PR #554, opened 2024-08-08
Permit yaml safe_load of aliases in automate ruby methods

jrafanie:
Psych 4 defaults to safe_load, which defaults to not permitting aliases or classes not in an approved list.

This is similar to what we do in the core application here: https://github.com/ManageIQ/manageiq/blob/46c992aaee664ea79713020e60c0342f703a8bc6/lib/extensions/yaml_load_aliases.rb#L9

The difference is we don't want to pull in application models/classes as permitted classes, at least until we know why we need them. Also, automate's ruby invocation is somewhat isolated from the application and doesn't really pull much into the remote ruby process beyond active support and some minor changes.

We're instead, just extending YAML.safe_load to permit aliases in this change. We can add more later or find a better way to share code if that is needed.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
@@ -110,6 +110,14 @@ begin
   require 'drb'
   require 'yaml'
 
+  YAML.singleton_class.prepend(
+    Module.new do
+      def safe_load(yaml, aliases: false, **kwargs)
+        super(yaml, aliases: true, **kwargs)
+      end
+    end
+  )
+
   Time.zone = 'UTC'
 
   MIQ_OK    = 0
--- a/spec/engine/miq_ae_method_spec.rb
+++ b/spec/engine/miq_ae_method_spec.rb
@@ -47,6 +47,38 @@ describe MiqAeEngine::MiqAeMethod do
       end
     end
 
+    context "with a script that tries to YAML.load with aliases" do
+      let(:script) do
+        <<-RUBY
+          YAML.load("---\na: &a\n  b: true  \n\ndevelopment:\n  <<: *a\n  c: false\n\n")
+        RUBY
+      end
+
+      it "logs and returns the correct exit status" do
+        allow($miq_ae_logger).to receive(:info).and_call_original
+        expect($miq_ae_logger).to receive(:info).with("Method exited with rc=MIQ_OK", :resource_id => 123).at_least(:once)
+        expect($miq_ae_logger).to_not receive(:error)
+
+        expect(subject).to eq(0)
+      end
+    end
+
+    context "with a script that tries to YAML.safe_load with aliases" do
+      let(:script) do
+        <<-RUBY
+          YAML.safe_load("---\na: &a\n  b: true  \n\ndevelopment:\n  <<: *a\n  c: false\n\n")
+        RUBY
+      end
+
+      it "logs and returns the correct exit status" do
+        allow($miq_ae_logger).to receive(:info).and_call_original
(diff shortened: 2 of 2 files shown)


lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -110,6 +110,14 @@ class AutomateMethodException < StandardError
   require 'drb'
   require 'yaml'
 
+  module AutomateYamlLoadAliases
+    def safe_load(yaml, aliases: false, **kwargs)
+      super(yaml, aliases: true, **kwargs)
```
Fryguy, line 115: Very minor, but I think we can use forwarding syntax now

```suggestion
    def safe_load(yaml, aliases: false, ...)
      super(yaml, aliases: true, ...)
```

jrafanie: ah yes!

Fryguy: Note I'm not 100% this format works - 2.7 implementations this wouldn't work, but I haven't tried with 3.0+

spec/engine/miq_ae_method_spec.rb
```diff
@@ -47,6 +47,22 @@ def root
       end
     end
 
+    context "with a script that tries to load yaml with aliases" do
+      let(:script) do
+        <<-RUBY
+          YAML.load("---\na: &a\n  b: true  \n\ndevelopment:\n  <<: *a\n  c: false\n\n")
```
Fryguy, line 53: Not really important, but to make the intention clearer, should we also have a test that calls safe_load (or perhaps just change this test to call safe_load)?  Up to you...`load` is more common, and we've overridden safe_load, so this does verify that `load` _also_ works in addition to `safe_load`

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
jrafanie: sad trombone!  doesn't work.  :-(. Jason thinks it's for forward all and not forward the rest of the args.

spec/engine/miq_ae_method_spec.rb
jrafanie: yes, we should have both tests, good idea... interfaces of load and safe_load changed over time so I'm fine with being cautious with testing it

lib/miq_automation_engine/engine/miq_ae_engine/drb_remote_invoker.rb
```diff
@@ -110,6 +110,14 @@ class AutomateMethodException < StandardError
   require 'drb'
   require 'yaml'
 
+  module AutomateYamlLoadAliases
```
Fryguy, line 113: Also super minor, but the only other thing I might do different is to just use an anonymous module here.  That way we can't possible "clobber" a user script that happens to have this module name (which is probably infinitely rare anyway)

Fryguy: Something like

```ruby
YAML.singleton_class.prepend(
  Module.new do
    def safe_load(yaml, aliases: false, **kwargs)
      super(yaml, aliases: true, **kwargs)
    end
  end
)

jrafanie: pushed, let me know if that's the direction you were thinking...

jrafanie: pushed again, rubocop liked both ways, so I went with yours

Fryguy (approved):

PR #545, opened 2024-04-15
Fix automation array values

GilbertCherrie:
Fixed automation array values for multi select drop downs in Order Service Forms.

When ordering from a Service Catalog for an Ansible Playbook Catalog Item there is a bug with multi-select dropdown values not being correctly sent to the Ansible Playbook.

An example order service form with a multi-select dropdown:
Screenshot 2024-04-22 at 3 38 28 PM

Before:
When viewing the Ansible Playbook output in the Services / My Services Screen for the Service you order will look like this. `test1` is the multi-select dropdown value and it outputs "VARIABLE IS NOT DEFINED!". `test2` is just a single select dropdown used for reference.
Screenshot 2024-04-22 at 3 38 11 PM

This is due to the field being with "Array::" on its key value causing the Ansible Playbook output to not be able to find that variable. When the "Array::" tag is removed from the key it then displays the Array as a String with `\u001f` separating the Array items.
Screenshot 2024-04-22 at 3 47 51 PM

After:
By removing the "Array::" tag from the variable key as well as converting the value being sent to the Ansible Playbook from a String to an Array we receive the Array in the Playbook and see it on the Playbook output correctly. This allows us to send Arrays to the Ansible Playbook and view it correctly on the output screen as individual items instead of as a String.
Screenshot 2024-04-22 at 3 30 43 PM

PR #555, opened 2024-08-15
Fix ae engine tags

GilbertCherrie:
Depends on: https://github.com/ManageIQ/manageiq-ui-classic/pull/9250 and https://github.com/ManageIQ/manageiq/pull/23155

Fixed ae engine tag field values after this pr changed how they work: https://github.com/ManageIQ/manageiq-automation_engine/pull/545.

Also, updated and added new cases to ensure this tag functionality is not accidentally broken in the future.

Screenshot 2024-08-15 at 3 27 17 PM

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine.rb
@@ -263,7 +263,7 @@ module MiqAeEngine
 
   def self.create_automation_attribute_array_value(value)
     value.collect do |obj|
-      obj.kind_of?(ActiveRecord::Base) ? obj.id.to_s : obj.to_s
+      obj.kind_of?(ActiveRecord::Base) ? "#{obj.class.name}::#{obj.id}" : obj.to_s
     end
   end
 
@@ -320,13 +320,22 @@ module MiqAeEngine
     array_objects.each do |array_object|
       # Each array attribute is tagged with Array:: before the attribute key unless it already starts with Array::
       array_attr_key = array_object
-      if !array_object.starts_with?("Array::")
+      if !ae_attrs[array_object].empty? && ae_attrs[array_object].first.to_s.split("::").first == "Classification"
+        array_attr_key = if array_object.starts_with?("Array::")
+                           "Tag#{array_object}"
+                         else
+                           "TagArray::#{array_object}"
+                         end
+      elsif !array_object.starts_with?("Array::")
         array_attr_key = "Array::#{array_object}"
       end
       ae_attrs[array_attr_key] = ae_attrs[array_object].collect do |obj|
         obj.kind_of?(ActiveRecord::Base) ? "#{obj.class.name}::#{obj.id}" : obj.to_s
       end.join("\x1F")
       if !array_object.starts_with?("Array::")
+        ae_attrs.delete(array_object)
+      end
+      if !array_object.starts_with?("TagArray::") && array_attr_key.starts_with?("TagArray::")
         ae_attrs.delete(array_object)
       end
     end
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
@@ -258,19 +258,34 @@ module MiqAeEngine
     end
 
     def process_args_array(args, args_key)
-      # process Array::servers => MiqServer::2,MiqServer::3,MiqServer::4
+      # Process Array::servers => MiqServer::2,MiqServer::3,MiqServer::4
+      # If no class seperator exists then return Array::servers => 2,3,4
       key = args_key.split(CLASS_SEPARATOR).last
       value = args.delete(args_key)
       args[key.downcase] = load_array_objects_from_string(value)
     end
 
     def attribute_is_array?(attr)
-      attr.to_s.downcase.starts_with?("array::")
+      attr.to_s.downcase.starts_with?("array::") && !attr.to_s.downcase.starts_with?("tagarray")
     end
 
     def process_args_attribute(args, args_key)
-      # process MiqServer::svr => 2
-      if args_key.include?(CLASS_SEPARATOR)
+      # Process TagArray::dialog_tags => <Classification ...>, <Classification ...>, <Classification ...>
+      if args_key.to_s.downcase.starts_with?("tagarray")
+        tags = []
+        args[args_key].split("\u001F").each do |tag|
+          # args_key: TagArray::dialog_param_test1
+          # args[args_key]: <Classification ...>, <Classification ...>, <Classification ...>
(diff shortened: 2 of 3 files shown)


GilbertCherrie:
@miq-bot assign @jrafanie 
@miq-bot add-label bug

GilbertCherrie:
@miq-bot add-reviewer @jrafanie

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -262,8 +262,15 @@ def self.create_automation_attribute_array_key(key)
...
-      obj.kind_of?(ActiveRecord::Base) ? obj.id.to_s : obj.to_s
+    # When array of length 1 just return the single object or id instead of an array
+    if value.length == 1
+      obj = value[0]
+      obj.kind_of?(ActiveRecord::Base) ? "#{obj.class.name}::#{obj.id}" : obj.to_s
+    else
+      # Return an array of object references or ids
+      value.collect do |obj|
+        obj.kind_of?(ActiveRecord::Base) ? "#{obj.class.name}::#{obj.id}" : obj.to_s
+      end
```
jrafanie, line 273: is this the same as:


```ruby
      result = value.collect do |obj|
        obj.kind_of?(ActiveRecord::Base) ? "#{obj.class.name}::#{obj.id}" : obj.to_s
      end
      result.length == 1 ? result.first : result
```

It looks like we want to do the same thing regardless if it's a 1 element array or more, but we want to return just the first result if it's a single element array.

jrafanie: Do we need to be concerned with that we still are missing the `\x1F` that was there previously:  https://github.com/ManageIQ/manageiq-automation_engine/pull/545

I don't understand why that was there.

Fryguy: The \x1F was first introduce for string arrays that could have _actual_ commas in the value.  We chose a character that is designed for being a field separate and should never actually be a string member (\x1F is the ASCII character for a "Unit separator")

GilbertCherrie: We don't need the \x1F in this case because we are not converting the array values into a string but instead keeping them in an actual array.

jrafanie: I think it's fine to preserve the prior behavior.  It's odd that both the automate backend side and the user's automation scripts need to have conditional logic to deal with "sometimes array of strings, sometimes a single string"... but that's how it was.

I think we can deduplicate our code by building an array in the same way, then just return the first element of the array in the case of a 1 length array.  See my comment above.  I think that preserves the logic you had there.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -320,7 +327,13 @@ def self.create_ae_attrs(attrs, name, vmdb_object, objects = [MiqServer.my_serve
     array_objects.each do |array_object|
       # Each array attribute is tagged with Array:: before the attribute key unless it already starts with Array::
       array_attr_key = array_object
-      if !array_object.starts_with?("Array::")
+      if !ae_attrs[array_object].empty? && ae_attrs[array_object].first.split("::").first == "Classification"
```
jrafanie, line 330: New line 326 hurts my brain but let's not change that right now...

326 guarantees this first conditional is an array, so the empty check is fine, but, the next one is a bit concerning.

Maybe something like this:

```ruby
if !ae_attrs[array_object].empty? && ae_attrs[array_object].first.to_s.split("::").first == "Classification"
```

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
```diff
@@ -258,19 +258,37 @@ def process_args_as_attributes(args = {})
...
+      if args_key.to_s.downcase.starts_with?("tagarray")
+        tags = []
+        args[args_key].split("\u001F").each do |tag|
+          # args_key: TagArray::dialog_param_test1
+          # args[args_key]: <Classification ...>, <Classification ...>, <Classification ...>
+          # tag: <Classification ...>
+          key, klass = get_key_name_and_klass_from_key(tag)
+          tags.push(MiqAeObject.convert_value_based_on_datatype(key, klass))
+          args["#{key}_id"] = tag if attribute_for_vmdb_object?(klass, args[args_key]) && !@attributes.key?(key)
+        end
```
jrafanie, line 283: I don't understand this part.  I'll need to revisit it.

spec/miq_ae_engine_spec.rb
```diff
@@ -434,16 +444,32 @@ def call_automate(obj_type, obj_id, open_url_task_id = nil)
...
+      result_arr = if hash["tags"].length == 1
+                     "Classification::#{hash["tags"][0].id}"
+                   else
+                     hash["tags"].collect { |h| "Classification::#{h.id}" }
+                   end
+      result = MiqAeEngine.create_automation_attributes(hash)
+      expect(MiqAeEngine.create_automation_attributes_string(hash)).to eq(result_str)
+      expect(result["tags"]).to eq(result_arr)
+    end
+
```
jrafanie, line 462: @GilbertCherrie Did we completely revert the test changes so it's testing as it was before?  This test looks new but others are changes.  Basically, it's hard to determine the changes we're doing on code pre #545.  I can't tell if you did a full revert of the prior changes in the repos and then made changes on top or we're keeping some of the changes from #545 and just fixing it to work with the original tests.

lib/miq_automation_engine/engine/miq_ae_engine.rb
```diff
@@ -329,6 +337,9 @@ def self.create_ae_attrs(attrs, name, vmdb_object, objects = [MiqServer.my_serve
       if !array_object.starts_with?("Array::")
         ae_attrs.delete(array_object)
       end
+      if !array_object.starts_with?("TagArray::") && array_attr_key.starts_with?("TagArray::")
+        ae_attrs.delete(array_object)
+      end
```
jrafanie, line 342: I'm not sure what's being deleted here.  It looks like it's removing non-array objects.  If we need to do all of those conditions, you can combine them:

```ruby
if !array_object.starts_with?("Array::") || (!array_object.starts_with?("TagArray::") && array_attr_key.starts_with?("TagArray::"))
  ae_attrs.delete(array_object)
end
```

jrafanie:
Sorry, this fell off my radar... was this fixed and then went stale?  Do you want to get it rebased @GilbertCherrie

GilbertCherrie:
@jrafanie I put all the needed work we need in this PR description: https://github.com/ManageIQ/manageiq-content/pull/773. We had some what left this work as we realized it was a very complicated fix and were scared of this potentially breaking other things so we left it as is for now

PR #558, opened 2024-09-24
Avoid Fixnum

kbrock:
`Fixnum` has been deprecated for a while, it has finally been dropped.

I am not totally sure of the implications here.
This will essentially change the wire protocol.

In theory, we could leave the code as-is.
The references to `Fixnum` are not ruby, but rather just in xml. and we are using `value.to_i` locally.

But grep wise, would be nice for `grep Fixnum` to come back empty.
Which is a non-trivial statement.
Very hard to manage a migration when there are false positives all around.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_object.rb
@@ -190,7 +190,7 @@ module MiqAeEngine
       case value.class.to_s
       when 'MiqAePassword'            then xml.Password(OPAQUE_PASSWORD)
       when 'String'                   then xml.String(value)
-      when 'Fixnum'                   then xml.Fixnum(value)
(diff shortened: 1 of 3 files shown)


Fryguy (approved):

Fryguy:
Is drb needed for fixnum?  I'm confused how that's related.

jrafanie:
> Is drb needed for fixnum? I'm confused how that's related.

Sounds like something that was seen while playing around with this branch(last commit):

```
  warning: drb.rb was loaded from the standard library,
  but will no longer be part of the default gems starting from Ruby 3.4.0.
  You can add drb to your Gemfile or gemspec to silence this warning.
```

Seems like something we can extract and will need to do elsewhere but maybe there aren't much drb usages left so maybe it's just this.  🤞

kbrock:
update:
- removing ruby 3.3 changes since this is good for all ruby implementations

PR #565, opened 2025-01-30
Expose attribute_aliases as attributes

kbrock:


spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
```diff
@@ -140,6 +140,39 @@
...
   end
+
+  context 'with a StoreObject service model' do
+    let(:model) { CloudObjectStoreObject }
+    let(:service_model) { MiqAeMethodService::MiqAeServiceCloudObjectStoreObject }
+
+    # NOTE: CloudObjectStoreObject.alias_attribute :name, :key
+    it "finds attributes aliases (without virtual attribute defined)" do
+      # if this shows true, find another model/attribute
+      expect(model.virtual_attribute?("name")).to eq(false)
```
Fryguy, line 151: I'm a little confused what this is trying to show. For CloudObjectStoreObject, both "key" and "name" show up in attribute_names, so I feel like this test would pass without the code changes above.  Same goes for Flavor.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
```diff
@@ -60,7 +60,7 @@ def self.ar_model_associations
 
     def self.expose_class_attributes(subclass)
       subclass.class_eval do
-        model.attribute_names.each do |attr|
+        (model.attribute_names | model.try(:attribute_aliases)&.keys).each do |attr|
```
jrafanie, line 63: Nice find on this location.  Does it make sense to have method that virtual attributes adds that provides this?

I see we have similar code here and https://github.com/ManageIQ/manageiq-api/pull/1278.  Perhaps, we could have something in core that extends AR to give us `exposed_attribute_names` so we can have one place to test that `exposed_attribute_names` includes virtual attributes, columns, aliased attributes, etc.  Perhaps deprecated_attributes are in the same bucket.

Fryguy: @jrafanie Did you see my comment above?

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
jrafanie: This is for rails 7.1.

```ruby
irb(main):002:0> Flavor.attribute_names.grep(/cpu/)
=> ["cpu_total_cores", "cpu_cores_per_socket", "cpu_sockets"]
irb(main):003:0> Flavor.attribute_aliases.keys.grep(/cpu/)
=> ["cpus", "cpu_cores"]
```

I *think* it's related to https://www.github.com/rails/rails/pull/52117 and the issue that caused it: https://www.github.com/rails/rails/issues/52111.  

Basically, we can't do `virtual_attribute` on things that are aliased attributes.  We're moving to just keeping `alias_attribute` and removing the `virtual_attribute` for aliased attributes.  Therefore, we need to change automate to pull in aliased attributes since we can't rely on attributes/attribute_names to have them.

lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb
jrafanie: see above

spec/engine/miq_ae_method_service/miq_ae_service_model_base_spec.rb
jrafanie: the idea is to drop virtual_column or virtual_attribute on things that are aliased attributes, like this commit:

https://github.com/ManageIQ/manageiq/pull/23225/commits/ae95f19dec89f56e497d93bbe0c0846b707d9b33

jrafanie: We hack it into [attribute_names for virtual columns](https://github.com/ManageIQ/activerecord-virtual_attributes/blob/eb351fc08a2f57a9553805ad0d1ff11d25a8d065/lib/active_record/virtual_attributes.rb#L104). This doesn't work with rails 7.1 for aliased attributes because they changed how they work in rails 7.1.

Fryguy: ok thanks - I couldn't understand the context - I knew it was Rails 7.1, but it didn't make sense.

kbrock: @Fryguy @jrafanie Do we want to leverage [ar_visible_attributes.rb](https://github.com/ManageIQ/manageiq/blob/master/lib/extensions/ar_visible_attribute.rb#L20) here?

Maybe introduce `hidden_and_non_hidden_attributes` or `all_attributes`?

kbrock: something like this? https://github.com/ManageIQ/manageiq/pull/23324

kbrock:
WIP: introducing more generic method for the concept of attributes and attribute aliases. After that is done, will change this code to leverage that.

kbrock:
update:
- using all_attribute_names

jrafanie (approved):

jrafanie:
Removed radjabov... we'll be doing rails 7.1 in spassky

PR #570, opened 2025-04-04
[WIP] Move resource id param to logger initialize

agrare:

### DIFF
--- a/lib/manageiq/automation_engine/logger.rb
+++ b/lib/manageiq/automation_engine/logger.rb
@@ -1,31 +1,32 @@
 module ManageIQ
   module AutomationEngine
     class Logger < ManageIQ::Loggers::Base
-      attr_reader :automation_log_wrapper
+      attr_reader :automation_log_wrapper, :resource_id
 
-      def initialize(*args, automation_log_wrapper:, **kwargs)
+      def initialize(*args, automation_log_wrapper:, resource_id: nil, **kwargs)
         @automation_log_wrapper = automation_log_wrapper
+        @resource_id            = resource_id
+
         super(*args, **kwargs)
       end
 
-      def self.create_log_wrapper(io = File::NULL)
+      def self.create_log_wrapper(io: File::NULL, automation_log_wrapper: nil, resource_id: nil)
         # We modify the interface of logger methods such as info/warn/etc. to allow the keyword argument
         # resource_id. Therefore, we need to wrap all client logger calls to these methods to process the resource_id,
         # cut the request_log entry and forward the remaining arguments to the logger.
-        new(io, :progname => "automation", :automation_log_wrapper => Vmdb::Loggers.create_logger("automation.log"))
+        new(io, :progname => "automation", :automation_log_wrapper => automation_log_wrapper || Vmdb::Loggers.create_logger("automation.log"), :resource_id => resource_id)
       end
 
-      private def add_to_db(severity, message = nil, progname = nil, resource_id: nil)
-        return [severity, message, progname] unless resource_id
+      private
 
-        # Adapted from Logger#add
-        severity ||= UNKNOWN
-        if severity < level
-          return [severity, message, progname]
-        end
-        if progname.nil?
-          progname = @progname
-        end
+      def add(severity, message = nil, progname = nil, &block)
+        automation_log_wrapper.add(severity, message, progname, &block)
+
+        severity ||= Logger::UNKNOWN
+        return true if severity < @level
+
+        progname ||= @progname
+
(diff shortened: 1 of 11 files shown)


lib/manageiq/automation_engine/logger.rb
```diff
@@ -1,31 +1,32 @@
...
 
-      private def add_to_db(severity, message = nil, progname = nil, resource_id: nil)
-        return [severity, message, progname] unless resource_id
+      private
+
+      def add(severity, message = nil, progname = nil, &block)
+        automation_log_wrapper.add(severity, message, progname, &block)
+
+        severity ||= Logger::UNKNOWN
+        return true if severity < @level
```
kbrock, line 26: Do we want our level, or the level of our wrapped log?

(that is the whole `automation_log_wrapper.send("#{severity}?")` thing)

lib/manageiq/automation_engine/logger.rb
```diff
@@ -1,31 +1,32 @@
...
-        new(io, :progname => "automation", :automation_log_wrapper => Vmdb::Loggers.create_logger("automation.log"))
+        new(io, :progname => "automation", :automation_log_wrapper => automation_log_wrapper || Vmdb::Loggers.create_logger("automation.log"), :resource_id => resource_id)
       end
 
-      private def add_to_db(severity, message = nil, progname = nil, resource_id: nil)
-        return [severity, message, progname] unless resource_id
+      private
+
+      def add(severity, message = nil, progname = nil, &block)
+        automation_log_wrapper.add(severity, message, progname, &block)
```
kbrock, line 23: think this wants to be after we resolve `message`?

Doesn't seem optimal if we `message ||= yield` and then have `automation_log_wrapper` have to do the same thing

kbrock (commented):
good stuff

lib/manageiq/automation_engine/logger.rb
agrare: I was thinking of the wrapped logger had a different progname that it wanted to use this would allow that

agrare: Plus I think this matches the existing behavior because it used to call the wrapped logger directly from `def info` as opposed to from `add`

lib/manageiq/automation_engine/logger.rb
agrare: Maybe we sync the `:level` from the wrapped logger when we build it initially?  I don't want to add _more_ call-outs to the 'hard-coded' log_wrapper since I want to try to drop this and just use a normal broadcast

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb
```diff
@@ -168,11 +168,13 @@ def self.with_automation_env
     private_class_method :with_automation_env
 
     def self.process_ruby_method_results(return_code, msg, miq_request_id)
+      logger = ManageIQ::AutomationEngine::Logger.create_log_wrapper(:resource_id => miq_request_id)
```
kbrock, line 171: I wonder if we can simplify the logging even further. Moving a logger into the workspace is helpful, but also thinking about creating the log wrapper being an even simpler method.

```ruby
ManageIQ::AutomationEngine.logger(miq_request_id)
```

That is just an example, haven't thought through this properly.

lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_workspace_runtime.rb
```diff
@@ -63,7 +67,7 @@ def self.instantiate_with_user(uri, user, attrs)
 
       workspace
     rescue MiqAeException => err
-      $miq_ae_logger.error(err.message, :resource_id => miq_request_id)
+      workspace.logger.error(err.message, :resource_id => miq_request_id)
```
kbrock, line 70: did the `workspace.logger` allow us to no longer pass `:resource_id`?

agrare: Correct, https://github.com/ManageIQ/manageiq-automation_engine/pull/570/files#diff-33fc5f7f8206b817c0ca3d4ad954ddb6c20cffd0aafc8c6b5c1966fb40804ecbR32
So we pass the resource_id during the initializer not on every log invocation

kbrock: so we can change this line?
```suggestion
      workspace.logger.error(err.message)
```

PR #584, opened 2026-03-18
Drop MiqLdap

Fryguy:
Part of ManageIQ/manageiq#21127

@jrafanie Please review. None of these methods are found in the org (and by extension are not found in manageiq-content).


Fryguy:
Note that if you git grep ldap, there is more here, but it's all in specs, and Im ok with leaving it because it's based around a hypothetical DOMAIN class in automate that represented ldap domains (not actually MiqLdap based)

jrafanie (approved):
✂️ 🔥 ✂️ 🔥 ✂️ 🔥 

🍰  🍪 👏 🙇 😍 🎉

PR #588, opened 2026-07-21
Replace eval in process_assertion with Prism-validated isolated eval

Fryguy:
Assertion values are fully substituted before evaluation — all ${...} tokens are resolved to literal values — so no Ruby variable names from the automation context are ever present in the evaluated string. The old eval(assertion) call ran in the method's own binding, unintentionally exposing MiqAeObject's local variables and self to user-supplied assertion strings.

Automate domain authoring is an admin-only task, so in practice this was only ever reachable by users who already have superuser access. This is not a security exposure, but it establishes a more principled foundation: assertions should be simple data comparisons, not arbitrary Ruby, and the implementation now reflects that intent.

This replaces the bare eval with two changes:

- MiqAeAssertion.valid? parses the assertion with Prism and walks the AST, rejecting anything outside a strict allowlist: literal values (booleans, nil, integers, floats, strings, symbols), comparison and logical operators, and Array#include? with a literal receiver and argument. Bare method calls (system, exit!), constant-qualified calls (File.read), backtick strings, assignments, and interpolated strings are all rejected before evaluation.

- MiqAeAssertion.evaluate runs the validated assertion against an ISOLATED_BINDING captured from a plain Object with no local variables, so eval cannot reach MiqAeObject internals even if validation were bypassed.

@jrafanie Please review.

---

## Public Repository Survey

To validate that the new `MiqAeAssertion` allowlist does not break any real-world automation content,
52 public ManageIQ/CloudForms automation repositories were surveyed for populated assertion field values.

### Methodology

For each repository, every `__class__.yaml` file was inspected to identify fields with `aetype: assertion`.
The corresponding instance YAML files were then checked to find any that actually set a non-empty value
on those fields.

### Results

Out of 52 repositories surveyed, the vast majority define assertion fields in their class schemas
(typically `guard`, `eval`, `vendor`, `placement`, or `sendmail`) but **no instance ever sets a value
on them** — the field is present as a schema slot but left blank in practice.

**5 instances across 3 repositories** had a non-empty assertion value set. These resolved to exactly
**2 distinct post-substitution expressions**:

| Post-substitution expression | Type | Passes new validator? |
|---|---|---|
| `"true"` / `"false"` | Boolean literal | ✅ Yes |
| `'vmware' == 'vmware'` | String equality comparison | ✅ Yes |

### Conclusion

No real-world assertion found in any public repository would be rejected by the new `MiqAeAssertion.valid?`
check. The allowlist covers 100% of observed usage while excluding constructs — bare method calls,
constant-qualified calls, backtick execution, string interpolation, assignments — that have no place
in an assertion expression.

### DIFF
--- a/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_assertion.rb
+++ b/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_assertion.rb
@@ -1,0 +1,142 @@
+require 'prism'
+
+module MiqAeEngine
+  # Validates and evaluates an assertion string (after substitution).
+  #
+  # Assertions must be simple Ruby expressions consisting only of literals,
+  # comparison/logical operators, and Array#include? calls with a literal
+  # receiver and literal argument.
+  #
+  # Any construct that could execute arbitrary code — bare method calls with
+  # no receiver (e.g. system(), exit!), constant-qualified calls (e.g.
+  # File.read), backtick strings, assignments, string interpolation, etc. —
+  # is rejected before evaluation.
+  module MiqAeAssertion
+    # Isolated binding with no local variables and a plain Object as self,
+    # so that eval cannot access MiqAeObject internals.
+    ISOLATED_BINDING = Object.new.instance_eval { binding }.freeze
+
+    # Literal leaf node types: the only values that may appear as operands.
+    LITERAL_NODE_TYPES = [
+      Prism::TrueNode,
+      Prism::FalseNode,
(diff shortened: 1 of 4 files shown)


Fryguy:
@miq-bot cross-repo-test manageiq-content

agrare (approved):

agrare:
LGTM, @Fryguy should we update the docs to make it clear what is allowed?

Fryguy:
Yeah that's a good idea.

